PR #13, opened 2023-03-29
Added login and visit deadline into Auth

fastfists:
Not sure what good values are to have, but its good to have
this PR here

- feat: Added login deadline and visit deadline


kmkhami:
We are probably fine just using the defaults here. Merge it! 👌🤞

PR #54, opened 2023-05-06
feature: create and save stripe customer on insert_user

kmkhami:


migrations/2023-05-06-205756_create_stripe_customers/up.sql
fastfists, line 1: Should we also put a users payment plan here too?

fastfists (commented):
Looks pretty good to me

migrations/2023-05-06-205756_create_stripe_customers/up.sql
kmkhami: Nah, we'll make a new table for that. 

It's best to keep table operations atomic. Each table should only be performing a single duty.

PR #66, opened 2023-05-10
65 if succeeds stripe hits out webhook

fastfists:
Not sure if all the arms needed are present but the basic idea is here.


kmkhami:
So I think the critical things to process on the webhook are: 

- [ ] CheckoutSessionCompleted
- [ ] CustomerCreated

We need customer created bc: 
image

Stripe only returns the customer_id without us having to make another network request for the sake of getting the email on that user_id which I don't want to do since it then requires that we make the `handle_webhook_query` function async

kmkhami:
We  **should** also store the id of the checkoutsession and payment intent for tracking purposes 

image

PR #109, opened 2023-05-28
Refactor cards to use postgresql instead of qdrant

fastfists:
- feature: Disallow similar cards
- migration: Added card_metadata and vote_metadata tables
- feat: Updated create_card and search_cards
- feat: get_card_by_id


kmkhami:
Rebased #109 on this

fastfists (commented):
Looks good, surprised how nasty the point id to card mapping got, but I don't see a better way to do it.

PR #130, opened 2023-06-08
feature: filtering based on oc_source_url

densumesh:


kmkhami:
Merge requirements

- [ ] Check that create_card works
- [ ] Check performance against prod db
- [ ] Check that search works against local db
- [ ] Check that search works against prod db

PR #147, opened 2023-06-16
139 feature allow users to delete cards they have created

densumesh:


densumesh:
needs to be rebased onto private card branch and then the CardCollision row for that card also has to be deleted

densumesh:
No, but I want to wait for you guys to approve the other pr before I do

PR #166, opened 2023-06-27
feature: route for downloading a file

skeptrunedev:


src/operators/file_operator.rs
```diff
@@ -317,7 +325,58 @@ pub async fn convert_docx_to_html_query(
...
+        .get_result(&mut conn)
+        .map_err(|_| DefaultError {
+            message: "Could not find file",
+        })?;
+
+    match file_metadata.private {
+        true => {
+            if user_uuid != file_metadata.user_id {
+                return Err(DefaultError {
+                    message: "File is private and you are not the owner",
```
densumesh, line 356: Prob should return a 403 Forbidden instead of a Default Error

src/handlers/file_handler.rs
```diff
@@ -102,3 +104,15 @@ pub async fn update_file_handler(
...
 }
+
+pub async fn get_file_handler(
+    file_id: web::Path<uuid::Uuid>,
+    pool: web::Data<Pool>,
+    user: LoggedUser,
+) -> Result<HttpResponse, actix_web::Error> {
+    let file = get_file_query(file_id.into_inner(), user.id, pool)
+        .await
+        .map_err(|e| ServiceError::BadRequest(e.message.to_string()))?;
```
densumesh, line 115: Same comment as below prob should be a 403 forbidden if perms are not correct

densumesh (approved):
LGTM

PR #170, opened 2023-06-27
bugfix: files can be downloaded by non-auth'ed users now, feature: route for deleting a file , feature: return file_name with file_id for `CardMetadata`'s

skeptrunedev:


densumesh (approved):
LGTM

PR #187, opened 2023-07-03
feature: use libreoffice to convert files

skeptrunedev:


src/card_parser.py
densumesh, line 1: Why do we need this python file? Are we calling it from rust?

densumesh (approved):
LGTM

PR #215, opened 2023-07-06
feature: upsert to verification table if data is a verification

cdxker:


src/handlers/verification_handler.rs
```diff
@@ -1,36 +1,74 @@
...
+                .link
+                .ok_or_else(|| ServiceError::BadRequest("No link on this card to verify".into()))?;
+
+            get_webpage_score(&link, &card.content).await?
+        }
+    };
+
+    if let VerifyData::CardVerification { card_uuid } = data {
+        // This is a vault call, so we need to update the card with the score
+        let _ = web::block(move || {
```
densumesh, line 66: should this not be awaited?

PR #269, opened 2023-07-12
feature: get private cards on main search

cdxker:


src/data/schema.rs
```diff
@@ -115,6 +115,7 @@ diesel::table! {
 diesel::table! {
     invitations (id) {
         id -> Uuid,
+        #[max_length = 100]
```
densumesh, line 118: Should these be here?

src/data/schema.rs
```diff
@@ -129,6 +130,7 @@ diesel::table! {
         topic_id -> Uuid,
         sort_order -> Int4,
         content -> Text,
+        #[max_length = 10]
```
densumesh, line 133: Same as above

densumesh (approved):
LGTM

PR #310, opened 2023-07-24
bugfix: paginate collections on user page

fastfists:


src/handlers/collection_handler.rs
```diff
@@ -59,20 +59,46 @@ pub async fn create_card_collection(
...
+            id: collection.id,
+            author_id: collection.author_id,
+            name: collection.name.clone(),
+            is_public: collection.is_public,
+            description: collection.description.clone(),
+            created_at: collection.created_at,
+            updated_at: collection.updated_at,
+            file_id: collection.file_id,
+        }).collect(),
+        page,
```
densumesh, line 99: do we need to return what page we are on?

PR #427, opened 2023-09-15
feature: use json column for metadata

densumesh:


server/src/operators/card_operator.rs
```diff
@@ -194,8 +197,24 @@ pub async fn search_card_query(
                 card_metadata_columns::author_id.eq(current_user_id.unwrap_or(uuid::Uuid::nil())),
             ))
             .distinct();
+    if !tag_set.is_empty() {
+        query = query
+            .filter(card_metadata_columns::tag_set.like(format!("%{}%", tag_set.get(0).unwrap())));
```
skeptrunedev: I think we should avoid the `unwrap`

server/src/operators/card_operator.rs
```diff
@@ -194,8 +197,24 @@ pub async fn search_card_query(
...
+    }
 
-    if let serde_json::Value::Object(obj) = &filters.unwrap() {
+    for tag in tag_set.iter().skip(1) {
+        query = query.or_filter(card_metadata_columns::tag_set.like(format!("%{}%", tag)));
+    }
+
+    if !link.is_empty() {
+        query =
+            query.filter(card_metadata_columns::link.like(format!("%{}%", link.get(0).unwrap())));
```
skeptrunedev: Some thing on `unwrap`

skeptrunedev (changes requested):
Generally think we should take another pass and remove the unwraps

skeptrunedev (approved):
Not in love with this, but need to move quickly. Outstanding bug being created in that filters are not adhere dto strictly.

PR #437, opened 2023-09-18
feature: api keys as an alternative auth mechanism

skeptrunedev:
image

### DIFF
--- a/server/src/handlers/register_handler.rs
+++ b/server/src/handlers/register_handler.rs
@@ -2,6 +2,7 @@ use actix_web::{web, HttpResponse};
 use argon2::{self, Config};
 use diesel::prelude::*;
 use once_cell::sync::Lazy;
+use rand::{distributions::Alphanumeric, Rng};
 use serde::{Deserialize, Serialize};
 
 use crate::{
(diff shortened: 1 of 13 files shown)


skeptrunedev:
Run the following in devtools console:

```
fetch("http://localhost:8090/api/user/set_api_key", {
    method: "GET",
    credentials: "include"
})
```

after signing in to get an API key. It will show up in the response on the devtools network tab.

server/src/operators/card_operator.rs
```diff
@@ -545,6 +545,7 @@ pub fn get_metadata_query(
                 user_columns::username,
                 user_columns::website,
                 user_columns::visible_email,
+                user_columns::api_key_hash,
```
densumesh, line 548: Do we want to include their api key in the return request? seems like it might be a security issue

densumesh, line 548: Do we want their api key in the return request? seems like a security issue

skeptrunedev: I mean the user has to be able to see their created api key some way or another.

fastfists (changes requested):

server/src/handlers/register_handler.rs
```diff
@@ -122,3 +123,14 @@ fn insert_user_from_invitation(
...
+
+pub fn generate_api_key() -> String {
+    let rng = rand::thread_rng();
+    let api_key: String = rng
+        .sample_iter(&Alphanumeric)
+        .take(32)
+        .map(char::from)
+        .collect();
+
+    api_key
```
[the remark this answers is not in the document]
skeptrunedev: Fixex

PR #439, opened 2023-09-18
cleanup: removed async stripe

### LINKED ISSUE #430: cleanup: remove everything to do with async_stripe in `server`
At some point it may be nice to add this back, but it is not needed for now. 

The business model has shifted away from SaaS such that this code is only making compilation take longer. 

completion requirement: `async_stripe` can be removed as a crate

RamK777-stack:
Removed async stripe

For now, I have commented out and tested it. do I need to remove those completely?

/closes #430 
/claim #430


RamK777-stack:
I have commented the lines having stripe dependency and tested. do I need to completely remove or okay for now?

densumesh:
Could you remove the files/lines which are no longer necessary?

skeptrunedev:
> I have commented the lines having stripe dependency and tested. do I need to completely remove or okay for now?

Completely remove for sure

fastfists (changes requested):
Can you completely remove the code instead of completely commenting them out.

RamK777-stack:
okay

RamK777-stack:
@fastfists Thanks !! I have removed those now.

PR #452, opened 2023-09-19
feature: make it easy to disable SSR and require auth for all routes

rishi-raj-jain:
fix https://github.com/arguflow/arguflow/issues/433
/claim https://github.com/arguflow/arguflow/issues/433

### DIFF
--- a/server/src/lib.rs
+++ b/server/src/lib.rs
@@ -149,10 +149,12 @@ pub async fn main() -> std::io::Result<()> {
                 web::scope("/api")
                     .service(
                         web::resource("/invitation")
+                            .wrap(auth_middleware.clone())
                             .route(web::post().to(handlers::invitation_handler::post_invitation)),
                     )
                     .service(
                         web::resource("/register/{invitation_id}")
+                            .wrap(auth_middleware.clone())
                             .route(web::post().to(handlers::register_handler::register_user)),
                     )
                     .service(
@@ -178,6 +180,7 @@ pub async fn main() -> std::io::Result<()> {
                     )
                     .service(
                         web::resource("/topic")
+                            .wrap(auth_middleware.clone())
                             .route(web::post().to(handlers::topic_handler::create_topic))
                             .route(web::delete().to(handlers::topic_handler::delete_topic))
                             .route(web::put().to(handlers::topic_handler::update_topic))
@@ -185,6 +188,7 @@ pub async fn main() -> std::io::Result<()> {
                     )
                     .service(
                         web::resource("/message")
+                            .wrap(auth_middleware.clone())
                             .route(
                                 web::post().to(
                                     handlers::message_handler::create_message_completion_handler,
@@ -197,7 +201,8 @@ pub async fn main() -> std::io::Result<()> {
                             ),
                     )
                     .service(
-                        web::resource("/messages/{messages_topic_id}").route(
+                        web::resource("/messages/{messages_topic_id}")
+                        .wrap(auth_middleware.clone()).route(
                             web::get().to(handlers::message_handler::get_all_topic_messages),
                         ),
                     )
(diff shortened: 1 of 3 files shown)


rishi-raj-jain:
@skeptrunedev how do I test the change? pardon the basic questions, but would love to get this done! Seems like I got the right approach at it?

rishi-raj-jain:
Left some routes because I think that might already require users to log in already.

rishi-raj-jain:
@skeptrunedev, @fastfists or @densumesh

How do I test the change? pardon the basic questions, but would love to get this done! Seems like I got the right approach at it?

fastfists:
Sure! you can have a look at our docs, we have a postman collection here https://docs.arguflow.ai/quickstart. Heres a [direct link](https://www.postman.com/interstellar-escape-79389/workspace/arguflow-api-reference/collection/28637863-c7833c30-5c19-427f-9cab-e8a4f961511e?action=share&creator=28637873) to the postman collection if your prefer that

rishi-raj-jain:
@fastfists I meant locally

rishi-raj-jain:
So what I added is the auth middleware to some of the routes, how do I build the rust project and run it locally?

fastfists:
Oh you would want to go into the server directory and call `cargo run`, this should rebuild the project. Using the Postman link sent change the Server 1 environment variable to `localhost:8090`

server/src/lib.rs
```diff
@@ -178,13 +180,15 @@ pub async fn main() -> std::io::Result<()> {
...
                         web::resource("/topic")
+                            .wrap(auth_middleware.clone())
                             .route(web::post().to(handlers::topic_handler::create_topic))
                             .route(web::delete().to(handlers::topic_handler::delete_topic))
                             .route(web::put().to(handlers::topic_handler::update_topic))
                             .route(web::get().to(handlers::topic_handler::get_all_topics)),
                     )
                     .service(
                         web::resource("/message")
+                            .wrap(auth_middleware.clone())
```
densumesh, line 191: Don't do this. In the Handler files, look at the params for the functions. The ones that require auth don't have an Option<LoggedUser>, but those that do just have LoggedUser. You need to write some code that can dynamically select between the two or edit the auth_handler code so that it has some way of switching between requiring auth and not.

rishi-raj-jain: @densumesh https://github.com/arguflow/arguflow/pull/452/files#diff-f76d67463607cfe631fabd8a20e7e0d9aae0e2dd507d7a4ec561a6095da63cde is what I'm using here, isn't that what you're saying?

skeptrunedev:
Going to close this for now, but please re-open when closer to completion. 

Instead of using `wrap`, please follow the pattern established with `LoggedUser`. 

This can be achieved by copying the pattern established in `auth_handler.rs` as follows: 

```
pub type LoggedUser = SlimUser;

impl FromRequest for LoggedUser {
    type Error = Error;
    type Future = Ready<Result<LoggedUser, Error>>;

    fn from_request(req: &HttpRequest, pl: &mut Payload) -> Self::Future {
        if let Ok(identity) = Identity::from_request(req, pl).into_inner() {
            if let Ok(user_json) = identity.id() {
                if let Ok(user) = serde_json::from_str(&user_json) {
                    return ready(Ok(user));
                }
            }
        }

        if let Some(authen_header) = req.headers().get("Authorization") {
            if let Ok(authen_header) = authen_header.to_str() {
                if let Some(pool) = req.app_data::<web::Data<Pool>>() {
                    if let Ok(user) = get_user_from_api_key_query(&authen_header.to_string(), pool)
                    {
                        return ready(Ok(user));
                    }
                }
            }
        }

        ready(Err(ServiceError::Unauthorized.into()))
    }
}
````

Then, at the beginning of a request, you can add a variable name with the type defined as shown below: 

```
pub async fn get_me(
    logged_user: LoggedUser,
    pool: web::Data<Pool>,
) -> Result<HttpResponse, actix_web::Error> {
```

The middleware will be automatically called by the `actix-web` framework.

PR #496, opened 2023-09-25
cleanup: remove all code related to verifications

Varun2002:


fastfists (changes requested):
Instead of removing the migration files, can you create a new migration that drops the verifications table

PR #538, opened 2023-09-30
bugfix: card closing after client side request

PrasathHari:
Bugfix for the issue https://github.com/arguflow/arguflow/issues/524.

Preventing Card expansion until both auth and card requests in client is finished.

### DIFF
--- a/search/src/components/CollectionPage.tsx
+++ b/search/src/components/CollectionPage.tsx
@@ -469,7 +469,7 @@ export const CollectionPage = (props: CollectionPageProps) => {
                 </Show>
                 <textarea
                   id="collection-query-textarea"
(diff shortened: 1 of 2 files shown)


skeptrunedev (changes requested):
I went into more detail about the desired way to do this here: https://github.com/arguflow/arguflow/issues/524#issuecomment-1741622129

The way you are fixing this is a bit sub-optimal. The request to get the cards returns w/ auth in context regardless. If auth is not ok then we want to allow expansion since we know that the cards request will not run. Then, when the card request returns, we want to allow expansion regardless. 

We should move the `setClientSideRequestFinished(true);` out of the second `if (response.ok)` logic branch in the second `createEffect` and allow it to run regardless of whether or not the response is ok. 

This is also true of the `CollectionPage`'s usage of `ScoreCardArray`

So the `CollectionPage` also needs changes.

skeptrunedev (approved):
Great fix @PrasathHari ! We really appreciate your contribution :heart: .

Issue #523, opened 2023-09-29
feature: allow for a `.env` variable that controls the lines available before `show more`


skeptrunedev:
### Description

Depending on the deployment, you may want more or less lines to show before `show more`. 

### Target(s)

`search`,`chat`

### Requirement to close

Merged PR that makes the lines before `show more` settable via a `.env` var.

### PR commit requirements for a bounty claim

1 commit with the following message: `feature: allow for a `.env` variable that controls the lines available before `show more``

### Community channels

Matrix is preferred. Reach out on discord or Matrix for further assistance. 

- [discord](https://discord.gg/CuJVfgZf54)
- [matrix](https://matrix.to/#/#arguflow-general:matrix.zerodao.gg)

Scorpil:
I can take a stab at this.

PR #543, opened 2023-09-30
feature: support for .env variable that controls how many lines are visible on collapsed cards

Scorpil:
Issue #523 @skeptrunedev


chat/src/components/ScoreCard.tsx
```diff
@@ -255,6 +262,7 @@ const ScoreCard = (props: ScoreCardProps) => {
           "text-ellipsis max-w-[100%] break-words space-y-5 leading-normal":
             true,
         }}
+        style={{"-webkit-line-clamp": linesBeforeShowMore}}
```
densumesh, line 265: We are using tailwind in this codebase, so please don't use the style param. Rather use the tailwind eqiviland and add it to the styles. i.e. https://tailwindcss.com/docs/line-clamp

search/src/components/ScoreCard.tsx
```diff
@@ -355,6 +362,7 @@ const ScoreCard = (props: ScoreCardProps) => {
             classList={{
               "line-clamp-4 gradient-mask-b-0": !expanded(),
             }}
+            style={{"-webkit-line-clamp": linesBeforeShowMore}}
```
densumesh, line 365: Same as above comment

search/src/components/ScoreCard.tsx
```diff
@@ -366,6 +374,7 @@ const ScoreCard = (props: ScoreCardProps) => {
               "text-ellipsis max-w-[100%] break-words space-y-5 leading-normal":
                 true,
             }}
+            style={{"-webkit-line-clamp": linesBeforeShowMore}}
```
densumesh, line 377: Same here

densumesh (changes requested):

densumesh:
You are on the right track, just use tailwind and this will be an easy approve! Please also provide screenshots of working functionality if you can!

Scorpil:
I opted in for style approach because you can't have dynamic values in tailwind classes, since CSS files are created statically at build time https://tailwindcss.com/docs/content-configuration#dynamic-class-names

It's possible to configure a number of classes being _always_ included, but there are 2 problems with that:

- unused bundle are still included in a final bundle which goes agains the tailwinds principles
- only pre-defined values are supported, i.e. if values 1 through 10 are included, only those env variable values will have an effect.

_Maybe_ it's possible to use dotenv to read env variable in tailwind config on a build time to include single class... This would mean rebuild is required for variable value change to take effect, but otherwise should work. I considered it and decided that added complexity is not worth it compared to "style" approach. But if you do not agree I can try implementing it this way.

densumesh:
> I opted in for style approach because you can't have dynamic values in tailwind classes, since CSS files are created statically at build time https://tailwindcss.com/docs/content-configuration#dynamic-class-names
> 
> It's possible to configure a number of classes being _always_ included, but there are 2 problems with that:
> 
> * unused classes are still included in a final bundle which goes agains the tailwinds principles
> * only pre-defined values are supported, i.e. if classes with values 1 through 10 are included, only those env variable values will have an effect.
> 
> _Maybe_ it's possible to use dotenv to read env variable in tailwind config on a build time to include single class... This would mean rebuild is required for variable value change to take effect, but otherwise should work. I considered it and decided that added complexity is not worth it compared to "style" approach. But if you do not agree I can try implementing it this way.

You actually bring up a good point. Did not think of it that way

skeptrunedev:
Reviewing this again. Agreed on your approach actually.

skeptrunedev (approved):
@Scorpil you actually crushed this issue, nice work man! Don't forget to leave a star if you haven't yet :heart: . 

I wound up catching another bug while reviewing w/ not having the gradient mask package on `chat` that I fixed. Then I also just ran prettier. No worries there, those things were not requirements in the issue. 

There is one thing I still don't like. We have to leave the tailwind `line-clamp` style since it is a utility class that adds more styles than just the lines - https://tailwindcss.com/docs/line-clamp - and I don't know. It just feels bad. Not quite sure how to fix it, but something to think about if you ever hit this in the future. Thanks for your work man!

PR #547, opened 2023-10-01
Configure "lines before 'read more'" value through Tailwind theme

Scorpil:

Related to #543 

Controlling Tailwind value from an .env variable is inherently tricky due to the way Tailwind works. I propose a simple alternative where the value is controlled centrally from Tailwind configuration instead.

@skeptrunedev let me know what you think, if it's not what you're looking for feel free to close the PR (it was quicker to open one than to try and explain this alternative approach in plain English).


skeptrunedev:
I like where you are thinking on this, but the problem is that we want it to be configurable via a `.env`. We don't want to `gitignore` the tailwind config such that you could change that value and still get updates. 

For now I am going to close this PR. However, I want to stress that we appreciate your contributions and hope we can find a different issue for you to tackle. Thanks! :heart:

Scorpil:
Fair enough, no problem at all.

PR #571, opened 2023-10-05
feature: add env flag for enabling download

skeptrunedev:


server/src/handlers/file_handler.rs
```diff
@@ -158,6 +158,13 @@ pub async fn get_file_handler(
     user: Option<LoggedUser>,
     _required_user: RequireAuth,
 ) -> Result<HttpResponse, actix_web::Error> {
+    let download_enabled = std::env::var("DOCUMENT_DOWNLOAD_FEATURE").unwrap_or("off".to_string());
```
densumesh, line 161: Don't we want to default to on?

skeptrunedev: Hmmm, yeah. Let me change it. Conflicted bc Match was concerned from a security perspective

skeptrunedev: Eh, changed my mind again. Leaving default off to reduce the odds of them messing up.

densumesh (approved):
LGTM

PR #578, opened 2023-10-06
Fix crash when PUBLIC_FILTER_ITEMS is not set

Scorpil:
Commit 690a1cedda82a77adf3f10e36f67cdc47656e52b introduced `PUBLIC_FILTER_ITEMS` environment variable, but it was never added to `.env.search`, leading to a crash on startup due to JSON.parse getting `undefined`.

This PR makes `PUBLIC_FILTER_ITEMS` optional and adds it to `.env.search` with a default value to serve as a usage example.

<sub>We probably need a better documentation around all the env vars...<sub>


Scorpil:
@skeptrunedev @fastfists any idea what's up with the lint job? No errors on my local env.

skeptrunedev (approved):
Thank you for this man. Much appreciated.

skeptrunedev:
> @skeptrunedev @fastfists any idea what's up with the lint job? No errors on my local env.

It's on us. We have a bug where the CI actions only work when a maintainer runs them.

PR #579, opened 2023-10-06
feature: add highlights to all searches and removed extranous model

densumesh:


skeptrunedev (approved):
This is sick! Highlights on RAG is insanely cool. 

image

PR #583, opened 2023-10-08
fix: change PUBLIC_SEARCH_URL to VITE_SEARCH_URL

Scorpil:
Fixing a minor mistake; this breaks a card link in docker setup (and issues a warning on docker-compose up).


skeptrunedev:
Thank you for the fix!

PR #586, opened 2023-10-09
feature: env for suggested queries

skeptrunedev:
image


Scorpil:
Shouldn't it have `PUBLIC_` prefix like the rest?

skeptrunedev:
> Shouldn't it have `PUBLIC_` prefix like the rest?

No, because this one is only needed server-side. 

There could be argument that it should be, but we don't want there to be a data leak if you have authRequired set to "on"

fastfists (approved):

PR #654, opened 2023-11-16
feature: vectara corpora logic on card level

fastfists:
Added AF-Dataset header for multiple datasets in the same server.


densumesh (approved):
LGTM

PR #660, opened 2023-11-18
ONNX model

densumesh:


docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: "3"
 
 services:
   db:
-    image: paradedb/paradedb:latest
+    image: postgres:15
```
fastfists, line 5: Shouldn't this be set to paradedb for the hybrid-search pr to work properly

densumesh: oh yeah good catch

PR #679, opened 2023-12-01
feature: qdrant point updating

densumesh:


server/src/handlers/card_handler.rs
```diff
@@ -606,9 +611,10 @@ pub async fn update_card_by_tracking_id(
         .await?
         .map_err(|_| ServiceError::BadRequest("Card not found".into()))?;
 
-    update_qdrant_point_private_query(
-        qdrant_point_id,
+    update_qdrant_point_query(
+        None,
```
skeptrunedev, line 615: I don't think this should be `None`

PR #688, opened 2023-12-05
feature: add support for organizations

skeptrunedev:


fastfists:
Looks good don't merge yet please. Changing my migrations currently and it would be weird to edit them out of order.

PR #697, opened 2023-12-09
feature: keycloak auth

densumesh:


skeptrunedev (approved):
Not 100% sure this is in a working state, but I think it's close enough

Merging to prevent conflicts later for the most part

PR #707, opened 2023-12-13
feature: add stripe customers table

skeptrunedev:


server/migrations/2023-12-13-045001_create-stripe-customers-table/up.sql
```diff
@@ -0,0 +1,14 @@
+-- Your SQL goes here
+ALTER TABLE
+    users
+ADD
+    CONSTRAINT users_email_unique UNIQUE (email);
+
+CREATE TABLE stripe_customers (
+    id UUID NOT NULL UNIQUE PRIMARY KEY,
+    stripe_id TEXT NOT NULL UNIQUE,
+    email TEXT NOT NULL UNIQUE,
```
densumesh, line 10: Do we need a constraint with the UNIQUE on this col alr

densumesh (approved):

server/migrations/2023-12-13-045001_create-stripe-customers-table/up.sql
skeptrunedev: The constraint was on the old users table

I think it's there alr, but I didn't want to check and figure it out so I just added one for sanity

PR #708, opened 2023-12-13
feature: tantivy recovery tool

densumesh:


skeptrunedev (approved):
Mildly surprised diesel is here, but seems fine in practice.

This looks right. Will have to test at some point. 

Also, once we add billing stuff we'll probably need logic to apply limits.

PR #709, opened 2023-12-13
feature: add BL to create stripe customer on callback

skeptrunedev:


server/src/operators/card_operator.rs
```diff
@@ -32,26 +32,11 @@ pub fn get_metadata_from_point_ids(
...
-            card_metadata_columns::updated_at,
-            card_metadata_columns::tag_set,
-            card_metadata_columns::card_html,
-            card_metadata_columns::private,
-            card_metadata_columns::metadata,
-            card_metadata_columns::tracking_id,
-            card_metadata_columns::time_stamp,
-            card_metadata_columns::dataset_id,
-        ))
+        .select(CardMetadata::as_select())
```
fastfists, line 39: Thats pretty cool

fastfists (approved):
Might need more bl but this is good start

PR #711, opened 2023-12-13
feature: partition qdrant by id rather than by collection

densumesh:


fastfists (approved):
LGTM we lowkey need to make a type for the payload for our sanity but its only in the qdrant operator so it hasn't spread its mess everywhere.

PR #712, opened 2023-12-14
feature: migrate pyo3 to candle

densumesh:


skeptrunedev (changes requested):
We should update requirements.txt b4 merging this

skeptrunedev (commented):
Looks good to me. 

Would still be good to adjust the requirements.txt file for the server-python tho

skeptrunedev (approved):
That's a beautiful reduction of python packages

PR #731, opened 2023-12-25
feature: client envs

densumesh:


fastfists (commented):
Does it make sense to separate configuration based on client or server? Why not just have both in the same struct?

PR #733, opened 2024-01-03
feature: move to a plan usage table

densumesh:


skeptrunedev (approved):
This is dope. Should be a major perf improvement

Issue #734, opened 2024-01-03
bugfix: new migration to modify update_files_storage_counts function to account for UPDATE case


skeptrunedev:
### Description

The below SQL trigger function does not account for the case when a file is updated. 

When a file is updated, if the size changes, then we must account for that change in this trigger. 

![image](https://github.com/arguflow/arguflow/assets/15804464/c741e415-6bf7-40b2-a76b-92410434a4bd)

### Target(s)

`server`

### Requirement to close

- [ ] PR contains commit named `bugfix: update_files_storage_counts function accounts for UPDATE case` and accurately fixes the issue 

### Community channels

Matrix is preferred. Reach out on discord or Matrix for further assistance. 

- [discord](https://discord.gg/CuJVfgZf54)
- [matrix](https://matrix.to/#/#arguflow-general:matrix.zerodao.gg)

Akshay-Patel-dev:
Can I try this one?

PR #737, opened 2024-01-04
bugfix: update_files_storage_counts function accounts for UPDATE case

jkcs:

@skeptrunedev 
fix: #734
/claim #734


skeptrunedev:
A couple of things:

Instead of modifying the existing migration, the idea is to do this in a new migration.

Also, I don't think this change will cause a trigger on UPDATE. I believe this just modifies what occurs in the delete case.

skeptrunedev (changes requested):
There are a few problems here where you are using `id` instead of `org_id` and I don't think the `INSERT` case should be getting changed at all from the old version. 

Also, heads up, a couple others have PRs now and we will accept the first fully correct one.

skeptrunedev:
Going to close because we merged #739

PR #739, opened 2024-01-04
bugfix: update_files_storage_counts function accounts for UPDATE case

Dhanus3133:
Solves #734
/claim #734


skeptrunedev (changes requested):
this seems good so far, but I believe that you also have to drop the old `update_files_storage_counts` function in addition to the trigger. 

After that, I'll test it and I think this is good to go.

Dhanus3133:
@skeptrunedev Just added the changes. Kindly have another look again thanks.

skeptrunedev (approved):
This looks good to me.

Great work @Dhanus3133 !

Also, please do star the repo if find the project interesting and want to follow it :heart: :star: .

PR #743, opened 2024-01-05
feature: Enforce unique organization names, generate random on conflict

fastfists:


skeptrunedev (approved):
Approving, but worth mentioning to take a second look at the update to make sure that it does a correct error message so someone using the API could kind of know what likely went wrong.

PR #746, opened 2024-01-06
feature: use actix middleware for dataset and logged user

densumesh:


skeptrunedev (approved):
this is really trick

absolutely huge improvement over the way things were previously happening 

pushed a final commit cleaning up the way the role got derived then fixed a clippy error

otherwise this looks great

PR #776, opened 2024-01-14
feature: chat ui working

fastfists:
Chat ui is fully working, some work is needed on the server side, issue with openAI streaming the events
will make issues for that


skeptrunedev (approved):
this is crazy

PR #808, opened 2024-01-20
feature: openrouter support

densumesh:


skeptrunedev (approved):
did not actually test this, but it looks right

some concern over our fork of openai_dive as well

PR #820, opened 2024-01-22
feature: add health checks to docker compose and isolate backend

qnlbnsl:
### TODO: Add health check for server and update dependencies for search and chat.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,19 +9,29 @@ services:
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: password
       POSTGRES_DB: vault
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U postgres"]
+      interval: 10s
+      timeout: 5s
+      retries: 10
     volumes:
       - pgdata:/var/lib/postgresql/data
-    ports:
-      - "5432:5432"
+    # ports:
+    #   - "5432:5432"
     networks:
       - app-network
 
   redis:
     image: redis:7.2.2
-    container_name: my-redis
+    container_name: redis
     restart: always
-    ports:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -9,19 +9,29 @@ services:
...
+      test: ["CMD-SHELL", "pg_isready -U postgres"]
+      interval: 10s
+      timeout: 5s
+      retries: 10
     volumes:
       - pgdata:/var/lib/postgresql/data
-    ports:
-      - "5432:5432"
+    # ports:
+    #   - "5432:5432"
```
skeptrunedev, line 20: why comment out all of the ports?

skeptrunedev (changes requested):

skeptrunedev (commented):
overall looks good besides the port removal

docker-compose.yml
qnlbnsl: They are not required. For development sure. however, in production all queries must go through your server. As they all exist on the same network access is guaranteed. Port mapping is only to expose it to the bridged network form docker.

qnlbnsl: Basically, left them there for people who may want to connect directly. Not needed otherwise

qnlbnsl: We could remove them and add to the documentation instead. Up to you guys

skeptrunedev: > Basically, left them there for people who may want to connect directly. Not needed otherwise

Hmm, I tend to think that we should just keep them for local dev convenience and let folks who plan to self-host modify it if they so choose.

qnlbnsl: done

skeptrunedev (approved):

PR #838, opened 2024-01-25
feature: added roles to api keys

densumesh:


fastfists:
Where do we guard read / writes?

I honestly think its a good idea to ship without these for now

fastfists (approved):
Looks good, little hacky for now but should work for mvp purposes

PR #843, opened 2024-01-26
feature: super slim dockerfile

fastfists:
Dockerfile.no-ocr is halved in size at 228MB and caches properly using
rust:1.75 instead of rust-nightly, hopefully will reduce the amount of full
builds


skeptrunedev (approved):
LGTM

PR #889, opened 2024-01-30
feature: collisions as a server env flag

fastfists:
- feature: collisions as envflag for create_chunk route
- feature: collisions as parameter at lowest level.
- feature: collision flag for everything else


skeptrunedev (approved):
make sure to test b4 merging

PR #894, opened 2024-01-31
feature: made highlights customizable and ability to turn them off

densumesh:


fastfists (changes requested):
This should be a request parameter not an DatasetEnv flag

fastfists (approved):

PR #909, opened 2024-02-02
feature: flag to delete chunks with files

densumesh:

### DIFF
--- a/server/src/handlers/file_handler.rs
+++ b/server/src/handlers/file_handler.rs
@@ -1,8 +1,6 @@
 use super::auth_handler::{AdminOnly, LoggedUser};
 use crate::{
-    data::models::{
-        DatasetAndOrgWithSubAndPlan, File, Pool, ServerDatasetConfiguration,
-    },
+    data::models::{DatasetAndOrgWithSubAndPlan, File, Pool, ServerDatasetConfiguration},
     errors::ServiceError,
     operators::{
         file_operator::{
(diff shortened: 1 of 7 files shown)


server/src/handlers/file_handler.rs
```diff
@@ -247,11 +245,18 @@ pub async fn get_user_files_handler(
 )]
 pub async fn delete_file_handler(
     file_id: web::Path<uuid::Uuid>,
+    delete_chunks: web::Query<Option<bool>>,
```
fastfists, line 248: Update utopia doc string

server/src/operators/dataset_operator.rs
```diff
@@ -251,7 +251,7 @@ pub async fn delete_dataset_by_id_query(
         .map_err(|_| ServiceError::BadRequest("Could not find files".to_string()))?;
 
     for file_id in file_ids {
-        delete_file_query(file_id, dataset.id, pool.clone())
+        delete_file_query(file_id, dataset.id, Some(false), pool.clone())
```
fastfists, line 254: Why wouldn't this be true?

fastfists (commented):
Left some questions

server/src/operators/dataset_operator.rs
densumesh: because we already deleted all cards associated with the dataset, could cause bugs if we try to redelete/ try to redelete

fastfists (approved):

PR #910, opened 2024-02-02
bugfix: chat UI problems such that it now works w/out issue

skeptrunedev:


fastfists (approved):
Looks good noting here that openrouter API key will need to be set on deployment

skeptrunedev:
> Looks good noting here that openrouter API key will need to be set on deployment

don't merge!!!!!!!!!

PR #921, opened 2024-02-08
feature: bypass pg completely when not using collisions

densumesh:

### DIFF
--- a/server/src/handlers/chunk_handler.rs
+++ b/server/src/handlers/chunk_handler.rs
@@ -45,7 +45,7 @@ pub struct CreateChunkData {
     pub tag_set: Option<Vec<String>>,
     /// File_uuid is the uuid of the file that the chunk is associated with. This is used to associate chunks with files. This is useful for when you want to delete a file and all of its associated chunks.
     pub file_id: Option<uuid::Uuid>,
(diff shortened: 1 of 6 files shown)


server/src/handlers/chunk_handler.rs
```diff
@@ -42,6 +45,8 @@ pub struct CreateChunkData {
     pub tag_set: Option<Vec<String>>,
     /// File_uuid is the uuid of the file that the chunk is associated with. This is used to associate chunks with files. This is useful for when you want to delete a file and all of its associated chunks.
     pub file_id: Option<uuid::Uuid>,
+    /// File_name is the name of the file that the chunk is associated with. This is used to associate chunks with files. This is useful for when you want to delete a file and all of its associated chunks.
+    pub file_name: Option<String>,
```
skeptrunedev, line 49: I am assuming that the docstring here needs to be changed?

skeptrunedev (changes requested):

skeptrunedev (changes requested):
after testing locally, i am not seeing any noticeable performance improvement which makes me believe that we probably shouldn't merge this

based on similar routes, it seems like we are only saving ~10ms with this change and I don't think that's worth the added complexity 

image

PR #922, opened 2024-02-08
bugfix: edit and view chunk pages

skeptrunedev:

### DIFF
--- a/search/src/components/ScoreChunk.tsx
+++ b/search/src/components/ScoreChunk.tsx
@@ -29,9 +29,9 @@ import { Tooltip } from "./Atoms/Tooltip";
 import { AiOutlineCopy } from "solid-icons/ai";
 import { FullScreenModal } from "./Atoms/FullScreenModal";
 import { useStore } from "@nanostores/solid";
-import { currentUser } from "../stores/userStore";
 import { clientConfig } from "../stores/envsStore";
 import { currentDataset } from "../stores/datasetStore";
+import { A } from "@solidjs/router";
 
 export const sanitzerOptions = {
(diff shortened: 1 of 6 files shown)


search/src/components/ScoreChunk.tsx
```diff
@@ -337,12 +332,7 @@ const ScoreChunk = (props: ScoreChunkProps) => {
                 }
                 tooltipText="Copy to clipboard"
               />
-              <Show
-                when={
-                  props.setOnDelete &&
-                  $currentUser()?.id == props.chunk.author?.id
-                }
-              >
+              <Show when={props.setOnDelete}>
```
densumesh, line 335: Can we grab the user role to check if they can delete?

search/src/components/ScoreChunk.tsx
```diff
@@ -354,11 +344,9 @@ const ScoreChunk = (props: ScoreChunkProps) => {
                   <FiTrash class="h-5 w-5" />
                 </button>
               </Show>
-              <Show when={$currentUser()?.id == props.chunk.author?.id}>
-                <a title="Edit" href={`/chunk/edit/${props.chunk.id}`}>
-                  <FiEdit class="h-5 w-5" />
-                </a>
-              </Show>
+              <A title="Edit" href={`/chunk/edit/${props.chunk.id}`}>
```
densumesh, line 347: Same with this one? Just check if they are admin or above

densumesh (changes requested):

search/src/components/ScoreChunk.tsx
skeptrunedev: Yes

densumesh (approved):

PR #931, opened 2024-02-10
feature: add more timer logs to search

fastfists:


server/src/handlers/chunk_handler.rs
```diff
@@ -725,6 +725,7 @@ pub async fn search_chunk(
     let parsed_query = parse_query(data.query.clone());
 
     let mut timer = Timer::new();
+    timer.add("Just got to handler");
```
skeptrunedev, line 728: I feel like we should make this more descriptive

skeptrunedev (changes requested):

PR #967, opened 2024-02-25
feature: delete organization route

skeptrunedev:


skeptrunedev:
next bug to solve here is that once all datasets are deleting, it seems like the user is deleted 

### Test cases 

1. Get user into multiple orgs 
2. Delete one org 
    - Check that other org still exists
3. Delete next org
4. Check that user still exists

fastfists (approved):

PR #1018, opened 2024-03-05
feature: add dataset config for collection name, qdrant url, and qdrant api key

densumesh:


server/src/data/models.rs
```diff
@@ -1,5 +1,7 @@
 #![allow(clippy::extra_unused_lifetimes)]
 
+use crate::get_env;
+
```
skeptrunedev, line 4:
```suggestion
use crate::get_env;
```

skeptrunedev (approved):

PR #1045, opened 2024-03-07
Ds/fix-qdrant-payload-size-error

densumesh:


server/migrations/2024-03-07-071217_drop-group-constraint/down.sql
```diff
@@ -0,0 +1 @@
+-- This file should undo anything in `up.sql`
```
fastfists, line 1: Need down migration pls

fastfists (approved):

skeptrunedev (approved):

PR #1087, opened 2024-03-15
feature: qdrant microservice to push valid points to qdrant in batch

fastfists:


skeptrunedev (changes requested):
Looks like a good PR, but let's not merge until we prove this will fix the bottleneck

fastfists:
Sadly this was not the bottleneck, closing as it causes too much complexity and breaks a lot of flexibility that the QDRANT_COLLECTION setting previously had.

PR #1101, opened 2024-03-20
feature: bulk push chunks into redis

fastfists:


server/src/handlers/chunk_handler.rs
```diff
@@ -74,9 +74,15 @@ pub struct CreateChunkData {
...
+#[derive(Serialize, Deserialize, Clone, ToSchema)]
+pub struct FailedChunk {
+    index: usize,
+    message: String,
+}
+
 #[derive(Serialize, Deserialize, Clone, ToSchema)]
 #[schema(example = json!({
-    "chunk_metadata": {
+    "chunk_metadata": [{
```
skeptrunedev, line 85: We should include both types in the examples imo.

server/src/handlers/chunk_handler.rs
```diff
@@ -45,7 +45,7 @@ use utoipa::ToSchema;
     "weight": 0.5,
     "split_avg": false
 }))]
-pub struct CreateChunkData {
+pub struct ChunkData {
```
skeptrunedev, line 48: I much prefer `CreateChunkData` to this name. Nit, but this makes it seem like this is the `model` instead of the Create request payload.

server/src/handlers/chunk_handler.rs
```diff
@@ -86,23 +92,59 @@ pub struct CreateChunkData {
...
+    "split_avg": false
+}))]
+pub enum CreateChunkData {
+    Single(ChunkData),
+    Batch(Vec<ChunkData>),
+}
+
+#[derive(Serialize, Deserialize, Clone, ToSchema)]
+#[serde(untagged)]
+pub enum SingleOrMultiChunkMetadata {}
```
skeptrunedev, line 146: Should this be empty?

server/src/handlers/chunk_handler.rs
```diff
@@ -114,7 +156,7 @@ pub struct UploadIngestionMessage {
     request_body(content = CreateChunkData, description = "JSON request payload to create a new chunk (chunk)", content_type = "application/json"),
     responses(
         (status = 200, description = "JSON response payload containing the created chunk", body = ReturnQueuedChunk),
-        (status = 400, description = "Service error relating to to creating a chunk, likely due to conflicting tracking_id", body = ErrorResponseBody),
+        (status = 400, description = "", body = ErrorResponseBody),
```
skeptrunedev, line 159: Why remove the description?

server/src/handlers/chunk_handler.rs
```diff
@@ -125,118 +167,157 @@ pub struct UploadIngestionMessage {
...
+        CreateChunkData::Batch(chunks) => chunks,
+    };
+
     let count_dataset_id = dataset_org_plan_sub.dataset.id;
 
+    let mut timer = Timer::new();
     let chunk_count = get_row_count_for_dataset_id_query(count_dataset_id, pool.clone())
         .await
         .map_err(|err| ServiceError::BadRequest(err.message.into()))?;
+    timer.add("get daataset count");
```
skeptrunedev, line 187:
```suggestion
    timer.add("get dataset count");
```

server/src/handlers/chunk_handler.rs
```diff
@@ -125,118 +167,157 @@ pub struct UploadIngestionMessage {
...
 
-    if chunk_count
+    if chunk_count + chunks.len()
```
skeptrunedev, line 189: Should just be `>` now I think

skeptrunedev (changes requested):

server/src/handlers/chunk_handler.rs
fastfists: Its kinda weird, this *is* the model now the CreateRequestPayload is the enum which can be one or the other

server/src/handlers/chunk_handler.rs
fastfists: oh forgot to delete that

server/src/handlers/chunk_handler.rs
fastfists: its just inaccurate.

server/src/handlers/chunk_handler.rs
skeptrunedev: Maybe call it `CreateSingleChunkData`? 

Or `CreateChunkData` and `CreateChunksData`?

skeptrunedev: This is exposed via the generated API clients

fastfists:
removed bulk_create_chunks

server/src/handlers/chunk_handler.rs
fastfists: Modified it, slighlty nastier code but better docs

skeptrunedev (approved):

PR #1106, opened 2024-03-21
Tf/test window env

tfolbrecht:
@skeptrunedev 
Here's one way of doing what we wanted without rebuilding with vite on a user's machine

it exposes variables by setting them in the window object and using them with an or in solid

node:18 does the building
nginx:1.25.4-alpine-slim is the runtime

The multi-stage build gets rid of node dependencies and extra steps.

The runetime container is 12.3MB with 1.2kb of it being the dist code 

## Note

needs nginx config and redirect rules
Not sure of behavior if VITE environment variables are also specified

### DIFF
--- a/search/.dockerignore
+++ b/search/.dockerignore
@@ -1,2 +1,3 @@
+/dist
 /node_modules
 .env
--- a/search/Dockerfile
+++ b/search/Dockerfile
@@ -1,11 +1,26 @@
-FROM node:18-alpine3.18 AS runtime
+FROM node:18 as build
+
 WORKDIR /app
 
-RUN apk update && apk add --no-cache openssl-dev libpq pkgconfig build-base
-COPY package.json yarn.lock ./
-RUN yarn install
+COPY package*.json ./
+
 COPY . .
-ENV HOST=0.0.0.0 \
-    PORT=3000
-EXPOSE 3000
-CMD yarn build && node ./dist/server/entry.mjs
+
+RUN yarn install
+RUN yarn run build
+
+FROM nginx:1.25.4-alpine-slim
+
+WORKDIR /usr/share/nginx/html
+
+COPY --from=build /app/dist /usr/share/nginx/html
+# below hack only seems to work if the container isn't built with corresponding vite variable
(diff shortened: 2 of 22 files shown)


tfolbrecht:
I might have missed some of place apiHost is set, inconsistent naming in some places

skeptrunedev:
With my PR, you also don't have to rebuild and it doesn't break VITE patterns.

search/Dockerfile
```diff
@@ -1,11 +1,33 @@
...
+
+RUN yarn install
+RUN yarn run build
+
+FROM nginx:1.25.4-alpine-slim
+
+WORKDIR /usr/share/nginx/html
+
+COPY --from=build /app/dist /usr/share/nginx/html
+# below hack only seems to work if the container isn't built with corresponding vite variable
```
skeptrunedev, line 17: I think using envsubst is significantly less hacky and the traditionally correct way to do this.

search/src/components/BookmarkPopover.tsx
```diff
@@ -31,7 +31,7 @@ export interface BookmarkPopoverProps {
 }
 
 const BookmarkPopover = (props: BookmarkPopoverProps) => {
-  const apiHost = import.meta.env.VITE_API_HOST as string;
+const apiHost = window.API_HOST || import.meta.env.VITE_API_HOST as string;
```
skeptrunedev, line 34: Fwiw, my PR literally does the samr thing, but it uses a VITE plug-in to do this instead of hard coding.

skeptrunedev (changes requested):
I think this is a worse approach than using vite-plugin-runtime-env and envsubst.

With envsubst, you also don't rebuild the image, you just use the traditional unix command to replace the ENV var in index.html.

search/.dockerignore
```diff
@@ -1,2 +1,3 @@
+/dist
```
skeptrunedev, line 1: This should probably be added to the other PR

skeptrunedev:
Closing in favor of #1104

PR #1113, opened 2024-03-21
feature: add request params for fetching events

skeptrunedev:


server/src/handlers/event_handler.rs
```diff
@@ -1,27 +1,43 @@
...
 };
 use actix_web::{web, HttpResponse};
 use serde::{Deserialize, Serialize};
+use utoipa::{schema, ToSchema};
+
+#[derive(Serialize, Deserialize, Debug, Clone, ToSchema)]
+#[schema(example = json!({
+    "page": 1,
+    "page_size": 10,
+    "type": "failed"
```
fastfists, line 15:
```suggestion
    "type": "card_action_failed"
```

fastfists (changes requested):
Only nit all else is good!

fastfists (approved):

PR #1118, opened 2024-03-22
bugfix: image subsystem for legal litigation use-cases

skeptrunedev:
- bugfix: image subsystem properly sets src for images
- bugfix: images for ImageModal using the dataset correctly
- bugfix: images and files for ScoreChunk and ImageModal


transfer-chunks.js
```diff
@@ -0,0 +1,40 @@
+const res = await fetch("https://api.trieve.ai/api/chunk/search", {
```
fastfists, line 1: Was this supposed to be added?

fastfists (changes requested):
Just a question about that one file.

transfer-chunks.js
skeptrunedev: Removing

fastfists (approved):

PR #1119, opened 2024-03-22
add containers configurable at runtime for chat & search

tfolbrecht:
What it says on the box, pls test out the docker containers 

``` sh
cd chat
docker build . -t nginx-chat
docker run -p 8080:80 --env VITE_API_HOST="http://localhost:8090/api" nginx-chat
```


fastfists (approved):
I had some isssues with the env's for the server, but I think its just my server env's were messed up, because it couldn't connect to qdrant.
 
Did confirm that it is able to connect to the server though which is more important.

Added commit to change the name from arguflow to trieve.

PR #1127, opened 2024-03-24
minor: prebake keycloak image with realm and theme

tfolbrecht:
It just copies configs into keycloak to make use more succulent

### DIFF
--- a/keycloak/Dockerfile
+++ b/keycloak/Dockerfile
@@ -1,0 +1,10 @@
+FROM quay.io/keycloak/keycloak:23.0.7
+
+WORKDIR /opt/keycloak
+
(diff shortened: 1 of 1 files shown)


keycloak/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM quay.io/keycloak/keycloak:23.0.7
+
+WORKDIR /opt/keycloak
+
+COPY ./themes/arguflow /opt/keycloak/themes/arguflow
```
skeptrunedev, line 5: Can we figure out how to change everything from `arguflow` to `trieve`?

skeptrunedev (changes requested):

skeptrunedev (approved):

PR #1128, opened 2024-03-24
visiblity: more logs when not able to send to qdrant

fastfists:

### DIFF
--- a/server/src/operators/qdrant_operator.rs
+++ b/server/src/operators/qdrant_operator.rs
@@ -287,8 +287,9 @@ pub async fn create_new_qdrant_point_query(
         .upsert_points_blocking(qdrant_collection, None, vec![point], None)
         .await
         .map_err(|err| {
-            log::info!("Failed inserting chunk to qdrant {:?}", err);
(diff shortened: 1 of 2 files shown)


server/src/operators/qdrant_operator.rs
```diff
@@ -288,7 +288,7 @@ pub async fn create_new_qdrant_point_query(
         .await
         .map_err(|err| {
             log::info!("Failed inserting chunk to qdrant {:?}", err);
```
skeptrunedev, line 290: Two things:

```suggestion
            log::error!("Failed inserting chunk to qdrant {:?}", err);
```

Then, can we also add a `sentry::capture_message(` here?

skeptrunedev (changes requested):

server/src/operators/qdrant_operator.rs
fastfists: I'll do you one better, added it so that it captures all errors when returning a `ServiceError` ( another point over `DefaultError`

tfolbrecht (commented):
Lgtm

tfolbrecht (approved):

skeptrunedev (approved):

PR #1133, opened 2024-03-25
cleanup: set search and chat analytics correctly

skeptrunedev:


tfolbrecht (commented):
Why not async instead of defer?

skeptrunedev:
> Why not async instead of defer?

dug up the 'ole twitter thread on this - https://x.com/csswizardry/status/1331721659498319873?s=20

tfolbrecht (approved):
Lgtm

PR #1135, opened 2024-03-25
ops: docker compose up config | wip: still needs embeddings params 

skeptrunedev:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -122,9 +122,7 @@ services:
       - 6334:6334
     
   splade-embeddings:
-    image: trieve/splade-embeddings
-    container_name: splade-embeddings
-    build: ./embedding-server/
+    image: trieve/splade-embeddings:demo
     ports:
       - 7070:7070
     environment:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -132,10 +132,7 @@ services:
       - API_KEY=${OPENAI_API_KEY}
 
   ingestion-service:
-    container_name: ingest 
-    build:
-      context: ./server/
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -111,76 +110,74 @@ services:
...
-      - OIDC_AUTH_REDIRECT_URL=${OIDC_AUTH_REDIRECT_URL}
-      - OIDC_ISSUER_URL=${OIDC_ISSUER_URL}
-      - GPU_SERVER_ORIGIN=${GPU_SERVER_ORIGIN}
-      - AWS_REGION=${AWS_REGION}
-      - SPARSE_SERVER_ORIGIN=${SPARSE_SERVER_ORIGIN}
-      - EMBEDDING_SERVER_ORIGIN=${EMBEDDING_SERVER_ORIGIN}
-      - RERANKER_SERVER_ORIGIN=${RERANKER_SERVER_ORIGIN}
-      - UNLIMITED=${UNLIMITED}
+      - HF_TOKEN=${HF_TOKEN}
+      - API_KEY=${OPENAI_API_KEY}
```
skeptrunedev, line 130: @fastfists why would we not replace this with TEI?

skeptrunedev (commented):
WIP: still need to figure out what we are doing for the embedding vectors here

docker-compose.yml
fastfists: We could but this is 3x simpler than using TEI, TEI requires 4 services, this is just 1.

skeptrunedev:
Will re-open at correct time once there is another thing to review `

PR #1139, opened 2024-03-25
feature: switch from deadpool_redis to bb8_redis

fastfists:
Moved off of deadpool_redis due to this clients having issues. Now using bb8


fastfists:
Also can switch to postgres from deadpool to bb8 if that is an issue with similar ease if that is an issue.

skeptrunedev (approved):

skeptrunedev (approved):

PR #1146, opened 2024-03-26
feature: migrate from redis to rabbitmq

densumesh:

### DIFF
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -122,6 +122,9 @@ postgres-openssl = "0.5.0"
 openssl = "0.10.64"
 utoipa-swagger-ui = { version = "6.0.0", features = ["actix-web"] }
 bb8-redis = "0.15.0"
+lapin = "2.3.1"
+deadpool = "0.10.0"
+deadpool-lapin = "0.11.0"
 
 
 [build-dependencies]
--- a/server/Cargo.toml
+++ b/server/Cargo.toml
@@ -123,7 +123,6 @@ openssl = "0.10.64"
 utoipa-swagger-ui = { version = "6.0.0", features = ["actix-web"] }
 bb8-redis = "0.15.0"
 lapin = "2.3.1"
-deadpool = "0.10.0"
 deadpool-lapin = "0.11.0"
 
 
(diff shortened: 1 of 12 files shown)


server/src/bin/ingestion-microservice.rs
```diff
@@ -130,47 +118,59 @@ fn main() {
...
+    rabbit_pool: actix_web::web::Data<models::RabbitPool>,
     web_pool: actix_web::web::Data<models::Pool>,
 ) {
     log::info!("Starting ingestion service thread");
 
-    let mut redis_connection = redis_pool
+    let rabbit_connection = rabbit_pool
         .get()
         .await
         .expect("Failed to fetch from redis pool");
```
fastfists, line 132:
```suggestion
        .expect("Failed to fetch from rabbit pool");
```

server/Cargo.toml
```diff
@@ -113,6 +113,9 @@ postgres-openssl = "0.5.0"
 openssl = "0.10.64"
 utoipa-swagger-ui = { version = "6.0.0", features = ["actix-web"] }
 bb8-redis = "0.15.0"
+lapin = "2.3.1"
+deadpool = "0.10.0"
```
fastfists, line 117:
```suggestion
```
Deadpool is not needed, deadpool_lapin replaces that

server/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 #[macro_use]
 extern crate diesel;
+use deadpool_lapin::{Manager, Pool};
```
fastfists, line 3: A bit of a nit but can these just be full scope pathed, we have so many Manager and Pool types that this could get confusing very fast

server/src/lib.rs
```diff
@@ -114,6 +116,26 @@ impl Modify for SecurityAddon {
...
 
+pub async fn set_up_rabbit() -> Pool {
+    let manager = Manager::new(get_env!("RABBITMQ_HOST", "RABBITMQ_HOST must be set"), ConnectionProperties::default());
+    let pool: Pool = deadpool::managed::Pool::builder(manager)
+        .max_size(10)
+        .build()
+        .expect("can create pool");
+
+    let channel = pool.get().await.expect("can get connection").create_channel().await.expect("can create channel");
+    let _ = channel
```
fastfists, line 127:
```suggestion
    channel
```
Let _ is only needed if you don't expect/unwrap/or return an error, in this case we expect so its not needed

fastfists (changes requested):
A few nits, tried not to be too picky though

skeptrunedev:
Honestly seems like this would work and be better than Redis, but closing for now due to complication of extra infra for existing deployments.

PR #1162, opened 2024-03-27
devops: compose up from mirror - no builds to deploy

tfolbrecht:
This PR modified the docker-compose.yaml to allow users to deploy the app without building.
It adds dockerfiles to pin out mirrored versions and custom entrypoints to simplify configuration.
It also assumes there's a top level .env with all necessary variables


docker-compose.yml
```diff
@@ -1,300 +1,198 @@
...
-      - AWS_REGION=${AWS_REGION}
-      - SPARSE_SERVER_ORIGIN=${SPARSE_SERVER_ORIGIN}
-      - EMBEDDING_SERVER_ORIGIN=${EMBEDDING_SERVER_ORIGIN}
-      - RERANKER_SERVER_ORIGIN=${RERANKER_SERVER_ORIGIN}
-      - UNLIMITED=${UNLIMITED}
-      - REDIS_CONNECTIONS=${REDIS_CONNECTIONS}
+      - HF_TOKEN=${HF_TOKEN}
+      - API_KEY=${OPENAI_API_KEY}
+    env_file:
+      .env
```
tfolbrecht, line 111: The splade embeddings and ingestion-service should be expanded out instead of reading .env

fastfists: Ingestion-service can be the exact same as the server, just annoying to have to double the size of the docker-compose.yml file

fastfists:
Can you rebase and squash a few of the badly named commits.

docker-compose.yml
fastfists: Splade-embeddings only needs 
```
HF_TOKEN
API_KEY
and SENTRY_URL
```

fastfists:
Also do we want to just have a single large .env.example file instead of having 3-4 different ones at this point. The example .env seems small and reasonable enough to just have comments of what each variable is for, could simply users lives.

fastfists (changes requested):
Seems to not be able to connect to keycloak. Set for re-review once it gets set up.

skeptrunedev:
Closing for now pending getting Sentry up

PR #1178, opened 2024-03-28
ops: rust checks to single action

fastfists:
Moved all the rust checks to a single action. The hope here is that the cargo cache being
cached between each step should result in faster total action speed


fastfists:
Meh, not much faster, and confuses reviewers. Closing

PR #1180, opened 2024-03-29
feature: return only ids from the requests

densumesh:

### DIFF
--- a/server/src/data/models.rs
+++ b/server/src/data/models.rs
@@ -3,6 +3,7 @@
 use crate::get_env;
 
 use super::schema::*;
+use crate::handlers::chunk_handler::ScoreChunkDTO;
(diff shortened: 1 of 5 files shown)


server/src/data/models.rs
```diff
@@ -390,6 +391,94 @@ pub struct ChunkMetadataWithFileData {
     pub weight: f64,
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone, Queryable)]
+pub struct ChunkMetadataIDs {
```
skeptrunedev, line 395:
```suggestion
pub struct SlimChunkMetadata {
```

skeptrunedev (changes requested):

skeptrunedev (approved):

PR #1222, opened 2024-04-03
feature: use quote words and negated words in qdrant

densumesh:


fastfists (approved):

fastfists (approved):
Thank GOD! finally using a standardized QdrantPayload Object. That gave me so much anxiety for so long.

PR #1232, opened 2024-04-04
feature: event filtering

densumesh:
image

### DIFF
--- a/server/src/data/models.rs
+++ b/server/src/data/models.rs
@@ -989,6 +989,7 @@ pub struct Event {
     pub event_data: serde_json::Value,
 }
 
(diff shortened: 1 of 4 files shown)


server/src/data/models.rs
```diff
@@ -989,6 +989,7 @@ pub struct Event {
     pub event_data: serde_json::Value,
 }
 
+//TODO: change these to Chunk instead of cards but need to rescroll pg to fix ids in events
```
fastfists, line 992: Would a migration script not be enough for this?

densumesh: Yes need to write a migration script to scroll PG

fastfists (approved):

PR #1262, opened 2024-04-10
bugfix: order on recommendations and group search

skeptrunedev:


server/src/handlers/group_handler.rs
```diff
@@ -1038,15 +1038,36 @@ pub async fn get_recommended_groups(
...
+    let recommended_chunk_metadatas = group_qdrant_query_result
+        .search_results
+        .into_iter()
+        .filter_map(|group| {
+            recommended_chunk_metadatas
+                .iter()
+                .find(|metadata| metadata.group_id == group.group_id)
+                .cloned()
+        })
+        .collect::<Vec<GroupScoreChunk>>();
```
densumesh, line 1070: Why do it twice?

densumesh (changes requested):

server/src/handlers/group_handler.rs
skeptrunedev: Lol, good catch

densumesh (approved):

PR #1268, opened 2024-04-11
feature: attempt of caching with s3

fastfists:


fastfists:
Tests complete! #1269 showed that using Dockerhub registry caching was better than s3 and GHA.

PR #1310, opened 2024-04-15
wip: bulk upload as default

fastfists:
The issue with this PR is that PostgreSQL `tracking_id` conflict will instantly hault all other operations within a bulk_upload.
This likely will need to have a way so that it bulk inserts / bulk reverts PostgreSQL transactions. Not sure the implications as
far as writting duplicate_tracking_id events would look like from that though.

Very unsure about this PR, will review more in the morning


fastfists:
SO Many crazy lines of logic. Closing for now. Maybe a future you will be able to solve it, but not today. Just going to do a Qdrant Backfill queue

PR #1311, opened 2024-04-15
feature: dataset tracking id

densumesh:


server/migrations/2024-04-15-171749_tracking-id-for-datasets/up.sql
```diff
@@ -0,0 +1,2 @@
+-- Your SQL goes here
+ALTER TABLE datasets ADD COLUMN tracking_id TEXT NULL;
```
skeptrunedev, line 2: Should be required unique between `tracking_id` and `organization_id`.

skeptrunedev, line 2: Needs an index

skeptrunedev (changes requested):
- Add a route for getting a dataset by `tracking_id` 
- Allow user to include `dataset_tracking_id` instead of `dataset_id` for `TR-Dataset` header

skeptrunedev (approved):

skeptrunedev (approved):

PR #1316, opened 2024-04-16
feature: add ability to be able to go directly to the search page for a dataset + feature: links to dashboard in search and chat

skeptrunedev:


search/src/components/Contexts/DatasetAndUserContext.tsx
```diff
@@ -157,7 +169,7 @@ export const DatasetAndUserContextWrapper = (
       const params = new URLSearchParams(window.location.search);
 
       params.set("organization", organization.id);
-      window.history.replaceState(
+      window.history.pushState(
```
densumesh, line 172: is there a reason to use push state instead of replace state?

skeptrunedev: Replace does not add a history record. It just replaces the current on. 

Push adds one which lets the back button work correctly

densumesh (approved):

PR #1329, opened 2024-04-17
feature: bulk upload operator

fastfists:
#1310 Is back with a vengance, bulk upload is working now, #1318 proved that we still had a bottleneck
at a postgres level, needing us to do bulk postgres writes, which would likely lead to a bottleneck at
an embedding vector level which would then require a vector embedding worker which would increase the
services by 3x which does not sound very fun.... so we are back here


server/src/bin/ingestion-worker.rs
```diff
@@ -333,25 +321,233 @@ async fn ingestion_worker(
...
+        // Split average or Collisions
+        for message in payload.ingestion_messages {
+            let upload_chunk_result = upload_chunk(message, web_pool.clone()).await;
+
+            if let Ok(chunk_uuid) = upload_chunk_result {
+                chunk_ids.push(chunk_uuid);
+            }
+        }
+        return Ok(chunk_ids);
+    }
```
densumesh, line 366: Should this not be ors between the conditions?

server/src/bin/ingestion-worker.rs
```diff
@@ -333,25 +321,233 @@ async fn ingestion_worker(
...
+    let all_content: Vec<String> = inserted_chunk_metadatas
+        .iter()
+        .map(|(chunk_metadata, _, _, _)| {
+            // handle
+            convert_html_to_text(
+                &chunk_metadata
+                    .chunk_html
+                    .clone()
+                    .unwrap_or(chunk_metadata.content.clone()),
+            )
```
densumesh, line 458: Cant we just the content field rather than converting to html again?

server/src/bin/ingestion-worker.rs
```diff
@@ -709,37 +911,60 @@ pub async fn readd_error_to_queue(
...
+        //                 tracking_id, error
+        //             ),
+        //         },
+        //     ),
+        //     web_pool.clone(),
+        // )
+        // .await
+        // .map_err(|err| {
+        //     log::error!("Failed to create event: {:?}", err);
+        // });
```
densumesh, line 932: Why is this no longer triggered?

server/src/operators/chunk_operator.rs
```diff
@@ -457,7 +566,7 @@ pub async fn insert_chunk_metadata_query(
     Ok(chunk_data)
 }
 
-// Reverts insertion event, DO NOT USE IF UPSERT == TRUE
+/// Bulk revert, assumes upsert chunk_ids were not upserted, only enterted
```
densumesh, line 569: What happens to the chunks if it gets reverted?

fastfists: It deletes the chunks it created, the reason why is because we have retries, if it retires to add it again it will fail because of `chunk_id`, `group_id`, and `file_id` collision.

server/src/bin/ingestion-worker.rs
fastfists: Good catch!, will update.

server/src/bin/ingestion-worker.rs
fastfists: #Yes we Can. Updating

densumesh (approved):

PR #1345, opened 2024-04-19
feature: only return points indexed in qdrant and return ErrorResponseBody for all error types

fastfists:


server/src/operators/qdrant_operator.rs
```diff
@@ -1125,3 +1126,30 @@ pub async fn get_point_count_qdrant_query(
...
+            None,
+            &[point_id.to_string().into()],
+            false.into(),
+            false.into(),
+            None,
+        )
+        .await
+        .map_err(|err| {
+            log::info!("Failed to count points from qdrant: {:?}", err);
+            ServiceError::BadRequest("Failed to count points from qdrant".to_string())
```
skeptrunedev, line 1151: Should correctly tag these prob

server/src/operators/qdrant_operator.rs
```diff
@@ -6,6 +6,7 @@ use crate::{
     handlers::chunk_handler::ChunkFilter,
 };
 use actix_web::web;
+use diesel::sql_types::Oid;
```
skeptrunedev, line 9: Is this used?

skeptrunedev (changes requested):

server/src/operators/qdrant_operator.rs
fastfists: oops rust analyzer auto imported

skeptrunedev (approved):

PR #1387, opened 2024-04-30
feature: add embedding query prefix to dashboard

densumesh:

### DIFF
--- a/dashboard/src/pages/Dashboard/Dataset/DatasetSettingsPage.tsx
+++ b/dashboard/src/pages/Dashboard/Dataset/DatasetSettingsPage.tsx
@@ -698,6 +698,10 @@ export const ServerSettingsForm = () => {
               >
                 Embedding Query Prefix
               </label>
+              <AiOutlineInfoCircle
+                class="h-5 w-5 text-neutral-400"
+                title="For some embedding models, the training data includes query prefixes. The default for Jina is 'Search for: '. You can experiment with different values."
+              />
               <input
                 type="text"
                 name="embeddingQueryPrefix"
--- a/dashboard/src/pages/Dashboard/Dataset/DatasetSettingsPage.tsx
+++ b/dashboard/src/pages/Dashboard/Dataset/DatasetSettingsPage.tsx
@@ -32,6 +32,7 @@ export const defaultServerEnvsConfiguration: ServerEnvsConfiguration = {
   LLM_BASE_URL: "",
   LLM_DEFAULT_MODEL: "",
   EMBEDDING_BASE_URL: "https://embedding.trieve.ai",
+  EMBEDDING_MODEL_NAME: "jina-base-en",
   MESSAGE_TO_QUERY_PROMPT: "",
   RAG_PROMPT: "",
   EMBEDDING_SIZE: 768,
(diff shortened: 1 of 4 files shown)


dashboard/src/pages/Dashboard/Dataset/DatasetSettingsPage.tsx
```diff
@@ -42,6 +42,7 @@ export const defaultServerEnvsConfiguration: ServerEnvsConfiguration = {
   COLLISIONS_ENABLED: false,
   FULLTEXT_ENABLED: true,
   QDRANT_COLLECTION_NAME: null,
+  EMBEDDING_QUERY_PREFIX: "",
```
skeptrunedev, line 45: Shouldn't set this to `Search for: `?

skeptrunedev (changes requested):

dashboard/src/pages/Dashboard/Dataset/DatasetSettingsPage.tsx
densumesh: Shouldn't matter bc the server has it set as a default

fastfists: Wait no, we should set `Search For: ` only if they choose the jina embedding model. If they use the openai model it could be messed up

skeptrunedev: But if you set it to empty it won't use the default right?

fastfists: If its empty, it should just be empty by default. The default ServerSetting should be empty

densumesh, line 45:
```suggestion
  EMBEDDING_QUERY_PREFIX: "Search for: ",
```

fastfists (changes requested):

dashboard/src/pages/Dashboard/Dataset/DatasetSettingsPage.tsx
skeptrunedev: > Wait no, we should set `Search For: ` only if they choose the jina embedding model. If they use the openai model it could be messed up

But then we need to backfill for other dataset settings that use Jina

dashboard/src/pages/Dashboard/Dataset/DatasetSettingsPage.tsx
```diff
@@ -681,6 +691,30 @@ export const ServerSettingsForm = () => {
               </select>
             </div>
 
+            <div class="col-span-4 space-y-1 sm:col-span-2">
+              <label
+                for="embeddingSize"
+                class="mr-2 block text-sm font-medium leading-6"
+              >
+                Embedding Query Prefix
```
skeptrunedev, line 699: Shouldn't this also only be editable on creation?

skeptrunedev (changes requested):

dashboard/src/pages/Dashboard/Dataset/DatasetSettingsPage.tsx
fastfists: Not really. Just any new ones. this query prefix only gets applied at search time right?

skeptrunedev (approved):

PR #1392, opened 2024-05-03
cleanup: add 404 responses to proper routes

skeptrunedev:
- bugfix: search UI location display
- cleanup: properly return 404 responses for resource get routes

### DIFF
--- a/server/src/operators/group_operator.rs
+++ b/server/src/operators/group_operator.rs
@@ -36,7 +36,11 @@ pub async fn get_group_from_tracking_id_query(
         .filter(chunk_group_columns::tracking_id.eq(tracking_id))
         .first::<ChunkGroup>(&mut conn)
         .await
-        .map_err(|_err| ServiceError::BadRequest("Group not found".to_string()))?;
+        .map_err(|err| {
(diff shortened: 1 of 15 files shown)


server/src/operators/group_operator.rs
```diff
@@ -36,7 +36,11 @@ pub async fn get_group_from_tracking_id_query(
         .filter(chunk_group_columns::tracking_id.eq(tracking_id))
         .first::<ChunkGroup>(&mut conn)
         .await
-        .map_err(|_err| ServiceError::BadRequest("Group not found".to_string()))?;
+        .map_err(|err| {
+            log::error!("Error getting group from tracking_id {:}", err);
+
+            ServiceError::NotFound
```
densumesh, line 42: Can you modify the service error so that it returns a message with the 404 so it's more descriptive?

densumesh (changes requested):

server/src/operators/group_operator.rs
skeptrunedev: Done

densumesh (approved):

PR #1405, opened 2024-05-07
feature: better query generation using RAG

densumesh:

### DIFF
--- a/server/src/handlers/message_handler.rs
+++ b/server/src/handlers/message_handler.rs
@@ -867,17 +867,19 @@ pub struct SuggestedQueriesResponse {
     )
 
 )]
-#[tracing::instrument]
+#[tracing::instrument(skip(pool))]
 pub async fn create_suggested_queries_handler(
     data: web::Json<SuggestedQueriesRequest>,
     dataset_org_plan_sub: DatasetAndOrgWithSubAndPlan,
+    pool: web::Data<Pool>,
     _required_user: LoggedUser,
 ) -> Result<HttpResponse, ServiceError> {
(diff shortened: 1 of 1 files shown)


server/src/handlers/message_handler.rs
```diff
@@ -895,12 +897,50 @@ pub async fn create_suggested_queries_handler(
...
+                "Doc {}: {}",
+                idx + 1,
+                chunk.metadata.first().unwrap().content.clone()
+            )
+        })
+        .collect::<Vec<String>>()
+        .join("\n\n");
+
+    let content = ChatMessageContent::Text(format!(
+        "Here is some context for the dataset for which the user is querying for {}. Generate 3 suggested keyword searches based off this query a user made. Your only response should be the 3 keyword searches which are seperated by new lines and are just text and you do not add any other context or information about the keyword searches. This should not be a list, so do not number each keyword search. These keyword searches should be related to the user query and within the domain of the dataset. Here is the user query: {}",
```
skeptrunedev, line 936:
```suggestion
    let content = ChatMessageContent::Text(format!(
        "Here is some context for the dataset for which the user is querying for {}. Generate 3 suggested keyword searches based off this query a user made. Your only response should be the 3 keyword searches which are seperated by new lines and are just text and you do not add any other context or information about the keyword searches. This should not be a list, so do not number each keyword search. These keyword searches should be related to the user query and within the domain of the dataset.",
```

skeptrunedev (changes requested):

skeptrunedev (approved):

PR #1458, opened 2024-05-14
bugfix: make must not filters work better

densumesh:

### DIFF
--- a/server/src/operators/search_operator.rs
+++ b/server/src/operators/search_operator.rs
@@ -181,8 +181,6 @@ pub async fn assemble_qdrant_filter(
         }
     }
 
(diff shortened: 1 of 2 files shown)


server/src/operators/search_operator.rs
```diff
@@ -181,6 +181,8 @@ pub async fn assemble_qdrant_filter(
         }
     }
 
+    println!("Filter: {:?}", filter);
```
skeptrunedev, line 184:
```suggestion
```

skeptrunedev (changes requested):

skeptrunedev (approved):

PR #1490, opened 2024-05-20
feature: add filter options to message creation

drew-harris:


server/src/handlers/chunk_handler.rs
```diff
@@ -934,6 +934,41 @@ impl Default for SearchChunkData {
     }
 }
 
+fn merge_search_chunk_data(
```
drew-harris, line 937: Is there a better way to merge structs with lots of Options?

skeptrunedev (approved):

PR #1551, opened 2024-06-05
Adding the ability to change the log level (RUST_LOG) from environment variables

jooni22:


server/src/lib.rs
```diff
@@ -360,18 +360,29 @@ pub fn main() -> std::io::Result<()> {
...
-                        .add_directive(tracing_subscriber::filter::LevelFilter::INFO.into()),
+                        .add_directive(level_filter.into()),
                 ),
             )
             .init();
 
         None
     };
-
+
```
skeptrunedev, line 385: I believe this is causing the format CI action to fail: 

`cargo fmt --manifest-path=./server/Cargo.toml -- --check`

skeptrunedev (changes requested):

skeptrunedev (approved):

PR #1553, opened 2024-06-05
Fixed metadata on search to display objects

vid277:
If an object is uploaded to metadata, the search on the frontend is able to correctly parse and display this object.  

image


vid277:
image
image

skeptrunedev (approved):

PR #1591, opened 2024-06-13
feat: move api keys on the dashboard to new tab

drew-harris:


skeptrunedev (approved):

skeptrunedev:
I added a feature to show the dataset's and org's an api key can access when they are present: 

image

PR #1623, opened 2024-06-17
feature: active search for search website

drew-harris:


skeptrunedev (changes requested):
I haven't gone through quite everything yet, but two things high level: 

- [ ] should use the appearance of the Search.tsx page instead of Home for the active search so things feel less cluttered
- [ ] The layout shift from scrollbar is really bad. The move might be set `min-h-[101vh]` or something like that so it's always there.

skeptrunedev (changes requested):
As far as I can tell the only major thing left is restoring URL state for all params. 

URL state is extremely important because it allows for instant duplication of problems. If a particular query + param combo is causing an issue, it is easy to send a link with exactly that so someone else can replicate. 

We can't lose that functionality with this change.

skeptrunedev (approved):

PR #1629, opened 2024-06-18
feature: empty datasets

aaryanpunia:

### DIFF
--- a/server/src/bin/delete-worker.rs
+++ b/server/src/bin/delete-worker.rs
@@ -12,7 +12,7 @@ use trieve_server::{
     errors::ServiceError,
     establish_connection, get_env,
     operators::{
-        dataset_operator::{delete_dataset_by_id_query, DeleteMessage},
+        dataset_operator::{delete_chunks_in_dataset, delete_dataset_by_id_query, DeleteMessage},
         organization_operator::{
             delete_actual_organization_query, get_soft_deleted_datasets_for_organization,
         },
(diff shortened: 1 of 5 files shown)


server/src/bin/delete-worker.rs
```diff
@@ -210,6 +210,28 @@ async fn delete_worker(
...
+                    log::info!(
+                        "Deleted all chunks for dataset: {:?}",
+                        delete_worker_message.dataset_id
+                    );
+                    continue;
+                }
+                Err(err) => {
+                    log::error!("Failed to delete all chunks for dataset: {:?}", err);
+                    continue;
+                }
```
skeptrunedev, line 231: If the response is an `Ok` then it's important to remove it from the processing queue like: 

```
let _ = redis::cmd("LREM")
                    .arg("delete_dataset_processing")
                    .arg(1)
                    .arg(serialized_message)
                    .query_async::<redis::aio::MultiplexedConnection, usize>(&mut *redis_connection)
                    .await;
```

If the response is an error, then you need to add it back to the queue: 

```
let _ = readd_error_to_queue(delete_worker_message, err, redis_pool.clone()).await;
```

skeptrunedev (changes requested):

server/src/bin/delete-worker.rs
skeptrunedev: Can reference the code slightly lower than this for how this should work.

server/src/operators/dataset_operator.rs
```diff
@@ -256,6 +288,12 @@ pub async fn delete_dataset_by_id_query(
             break;
         }
 
+        let old_count = chunk_metadata_columns::chunk_metadata
+            .count()
+            .get_result::<i64>(&mut conn)
+            .await
+            .unwrap();
+
```
skeptrunedev, line 296: I think this and the following count should be removed?

skeptrunedev (changes requested):

server/src/operators/dataset_operator.rs
aaryanpunia: ah, residue from the logging I was doing!

server/src/operators/dataset_operator.rs
```diff
@@ -190,6 +191,37 @@ pub async fn soft_delete_dataset_by_id_query(
...
+    let mut redis_conn = redis_pool
+        .get()
+        .await
+        .map_err(|err| ServiceError::BadRequest(err.to_string()))?;
+
+    let message = DeleteMessage {
+        dataset_id: id,
+        server_config: config,
+        attempt_number: 0,
+        empty_database: true,
```
skeptrunedev, line 224:
```suggestion
        empty_dataset: true,
```

Nit

skeptrunedev (approved):

PR #1645, opened 2024-06-21
bugfix: ingestion workers only use 1 thread.

cdxker:
With kubernetes it is a lot easier to use multiple processes. This should improve speed by a lot.

It is important to note that the number of `REDIS_CONNECTIONS` should be reduced. We were using redis in the ingestion workers a lot more than we are now. Setting it to 2 seems decent for extra robustness without being excessive.


skeptrunedev (approved):

skeptrunedev (changes requested):
`thread_num` needs to be removed in all workers

skeptrunedev (approved):

PR #1651, opened 2024-06-24
feature: working TS monorepo

drew-harris:


drew-harris:
Eslint and prettier are fixed but still decoupled from each other (each package has its own eslint config)
Main benefit now is ability to share code!

skeptrunedev (approved):

PR #1654, opened 2024-06-25
enhancement: use future::join with multiple reranker calls

vid277:
- Uses `ureq` if there are fewer than 20 `request_docs`
- Creates async reqwest calls using `futures::join` if there are more than 20 `request_docs`

### DIFF
--- a/server/src/operators/model_operator.rs
+++ b/server/src/operators/model_operator.rs
@@ -742,53 +742,45 @@ pub async fn cross_encoder(
             results.index_mut(pair.index).score = pair.score as f64;
         });
     } else {
-        let vec_futures: Vec<_> = request_docs
+    let vec_futures: Vec<_> = request_docs
             .chunks(20)
             .enumerate()
             .map(|(i, docs_chunk)| {
-                let clipped_docs = docs_chunk
-                    .iter()
-                    .map(|doc| {
-                        if doc.len() > 7000 {
-                            doc.chars().take(20000).collect()
-                        } else {
-                            doc.clone()
-                        }
-                    })
-                    .collect::<Vec<String>>();
 
                 let parameters = CrossEncoderData {
                     query: query.clone(),
-                    texts: clipped_docs,
+                    texts: docs_chunk.iter().map(|s| s.to_string()).collect(),
                     truncate: true,
                 };
 
                 let cur_client = reqwest_client.clone();
-                let url = embedding_server_call.clone();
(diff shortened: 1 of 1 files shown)


server/src/operators/model_operator.rs
```diff
@@ -743,6 +741,77 @@ pub async fn cross_encoder(
...
+                let clipped_docs = docs_chunk
+                    .iter()
+                    .map(|doc| {
+                        if doc.len() > 7000 {
+                            doc.chars().take(20000).collect()
+                        } else {
+                            doc.clone()
+                        }
+                    })
+                    .collect::<Vec<String>>();
```
skeptrunedev, line 758:
```suggestion
```

server/src/operators/model_operator.rs
```diff
@@ -743,6 +741,77 @@ pub async fn cross_encoder(
...
+                            )
+                        })?;
+
+                    let embeddings: Vec<ScorePair> = serde_json::from_str(&embeddings_resp)
+                        .map_err(|e| {
+                            log::error!("Failed to format response from embeddings server {:?}", e);
+                            ServiceError::InternalServerError(
+                                "Failed to format response from embeddings server".to_owned(),
+                            )
+                        })?;
```
skeptrunedev, line 798: This request does not match the ureq request in terms of format. One specific example would be that it should be `.into_json::<Vec<ScorePair>>()` instead of `text()`.

server/src/operators/model_operator.rs
```diff
@@ -743,6 +741,77 @@ pub async fn cross_encoder(
...
+        let all_chunk_scores: Vec<(usize, Vec<ScorePair>)> = futures::future::join_all(vec_futures)
+            .await
+            .into_iter()
+            .collect::<Result<Vec<(usize, Vec<ScorePair>)>, ServiceError>>()?;
+
+        for (_index, chunk_scores) in all_chunk_scores {
+            for pair in chunk_scores {
+                results.index_mut(pair.index).score = pair.score as f64;
+            }
+        }
```
skeptrunedev, line 814: The `_index` here needs to be used to preserve the order. In this configuration, scores could be assigned at incorrect indices.

skeptrunedev (changes requested):

skeptrunedev:
image
image


Can replicate this with `trieve example` on the philosophize-this dataset. 

Something seems wrong. The first result for `page_size = 10` and `page_size > 10` should be the same order assuming that the first result is present in the set of re'rank'ed items in both cases as it is here.

PR #1656, opened 2024-06-25
bugfix: fix unwrap error with "ids or tracking ids" must be provided.

cdxker:

### DIFF
--- a/server/src/operators/search_operator.rs
+++ b/server/src/operators/search_operator.rs
@@ -166,9 +166,12 @@ pub async fn assemble_qdrant_filter(
                         }
                     }
                     ConditionType::HasID(cond) => {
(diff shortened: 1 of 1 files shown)


skeptrunedev (changes requested):
If a user passes in an ID which we don't have then I think we should tell them we don't have it instead of silently failing to add the filter.

cdxker:
Trying to push out the change faster, currently its crashing the server. So it silently fails AND 502s. This error occurs because a tracking id or a id *isn't* provided. The expected behavior if I do a filter with no id is for the filter to not be applied.

skeptrunedev:
Closing in favor of a new PR to target the unwrap

PR #1694, opened 2024-06-28
feature: use exact: false and indexed_only: true with qdrant queries

densumesh:


skeptrunedev (changes requested):
This needs to be a req param on the handler such that users don't think their data isn't ingested in the cloud.

Something to explicitly set at search req time.

skeptrunedev:
`exact: false` does not need to be to clarify

skeptrunedev:
Update: we ended up placing `indexed_only` into the `ServerDatasetConfiguration` so as to not create an additional request param for search routes.

skeptrunedev (approved):

PR #1702, opened 2024-07-01
feature: interpolate dataset id in code snippet and add copy button

drew-harris:


skeptrunedev:
image

skeptrunedev (changes requested):
After doing a fresh `yarn`, I could not render the dashboard due to a package issue. I think that `shiki` is missing.

dashboard/src/components/Codeblock.tsx
```diff
@@ -0,0 +1,52 @@
+/* eslint-disable solid/no-innerhtml */
+import { codeToHtml } from "shiki";
```
skeptrunedev, line 2:
```suggestion
import { codeToHtml } from "shiki";
```

This line specifically is causing the issue with shiki

skeptrunedev (changes requested):

skeptrunedev:
image

skeptrunedev (approved):

PR #1730, opened 2024-07-02
feature: add logging to delete worker

densumesh:

### DIFF
--- a/server/src/bin/delete-worker.rs
+++ b/server/src/bin/delete-worker.rs
@@ -216,12 +216,8 @@ async fn delete_worker(
         let delete_worker_message: DeleteMessage =
             serde_json::from_str(&serialized_message).expect("Failed to parse file message");
 
-        log::info!(
-            "Deleting chunks for dataset {:?}",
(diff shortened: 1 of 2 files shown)


skeptrunedev (changes requested):
Should also add logs during deletion. 

`"starting to clean dataset {}", id` or `"starting to delete dataset {}", id`
`"deleted 5,000 chunks from dataset {}", id`

server/src/bin/delete-worker.rs
```diff
@@ -216,6 +216,11 @@ async fn delete_worker(
         let delete_worker_message: DeleteMessage =
             serde_json::from_str(&serialized_message).expect("Failed to parse file message");
 
+        log::info!(
+            "Deleting chunks for dataset {:?}",
+            delete_worker_message.dataset_id
+        );
```
skeptrunedev, line 222: There should also be a log for clean

skeptrunedev (changes requested):

skeptrunedev (approved):

PR #1743, opened 2024-07-03
cleanup: change location to location_bias

densumesh:

### DIFF
--- a/server/src/handlers/chunk_handler.rs
+++ b/server/src/handlers/chunk_handler.rs
@@ -894,7 +894,7 @@ pub struct ChunkFilter {
             }
         ]
     },
-    "recency_bias": 1.0,
+    "date_bias": true,
     "use_weights": true,
     "get_collisions": true,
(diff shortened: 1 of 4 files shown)


server/src/handlers/chunk_handler.rs
```diff
@@ -894,7 +894,7 @@ pub struct ChunkFilter {
             }
         ]
     },
-    "recency_bias": 1.0,
+    "date_bias": true,
```
skeptrunedev, line 897:
```suggestion
    "recency_bias": 1.0,
```

server/src/handlers/chunk_handler.rs
```diff
@@ -1181,7 +1181,7 @@ pub async fn search_chunks(
...
```
skeptrunedev, line 1184:
```suggestion
    "recency_bias": 1.0,
```

skeptrunedev (approved):

PR #1749, opened 2024-07-03
feature: refactor ChunkGroup response to be ChunkGroupAndFileId in all handlers

aaryanpunia:


server/src/data/models.rs
```diff
@@ -1192,6 +1192,37 @@ pub struct ChunkGroupAndFile {
     pub updated_at: chrono::NaiveDateTime,
 }
 
+impl ChunkGroupAndFile {
```
skeptrunedev, line 1195:
```suggestion
impl ChunkGroupAndFileId {
```

skeptrunedev (changes requested):
As mentioned in the original issue since this only includes the `fileId`, we want to rename the model to `ChunkGroupAndFileId`

skeptrunedev (approved):

PR #1757, opened 2024-07-03
cleanup: remove collisions

densumesh:

### DIFF
--- a/server/migrations/2024-07-03-200355_remove-collisions/up.sql
+++ b/server/migrations/2024-07-03-200355_remove-collisions/up.sql
@@ -1,6 +1,5 @@
 -- Your SQL goes here
-UPDATE chunk_metadata
-SET qdrant_point_id = gen_random_uuid() -- Or any default value that makes sense
+DELETE FROM chunk_metadata
 WHERE qdrant_point_id IS NULL;
 
 -- Then, alter the table to make the column not nullable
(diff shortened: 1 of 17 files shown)


server/migrations/2024-07-03-200355_remove-collisions/up.sql
```diff
@@ -0,0 +1,11 @@
+-- Your SQL goes here
+UPDATE chunk_metadata
+SET qdrant_point_id = gen_random_uuid() -- Or any default value that makes sense
+WHERE qdrant_point_id IS NULL;
```
skeptrunedev, line 4: All rows where `qdrant_point_id` is NULL actually need to get dropped

skeptrunedev (changes requested):

server/src/operators/search_operator.rs
```diff
@@ -1370,16 +1255,8 @@ pub async fn retrieve_chunks_from_point_ids(
...
-                .clone()
-                .into_iter()
-                .filter(|chunk| chunk.metadata().qdrant_point_id == Some(search_result.point_id))
-                .collect();
-
-            collided_chunks.insert(0, chunk);
-
             ScoreChunkDTO {
-                metadata: collided_chunks,
+                metadata: vec![chunk],
```
skeptrunedev, line 1259: Pain :cry:

skeptrunedev (approved):

PR #1763, opened 2024-07-03
fix: Update README.md of docker setup

varshith257:
@skeptrunedev


skeptrunedev:
This is actually correct. Newer version of compose is `docker compose` and not `docker-compose`.

PR #1776, opened 2024-07-04
feature: markdown in chat UI

aaryanpunia:


skeptrunedev:
image

skeptrunedev (approved):

skeptrunedev:
There is a small regression on this where the citations are no longer links which can scroll the right chunk into view, but I think it's acceptable. 

We should change the default prompt to not include the citations though.

PR #1781, opened 2024-07-04
feature: full scan query

aaryanpunia:


skeptrunedev (changes requested):
Implementing these changes as I review, but this does not work. 

Cannot use a `count` query against Qdrant as that won't take the score threshold into account. 

Need to conduct search queries unfortunately :( .

skeptrunedev (approved):

PR #1782, opened 2024-07-04
bugfix: selected org/dataset should be respected when navigating to playgrounds

aaryanpunia:


chat/src/components/contexts/UserContext.tsx
```diff
@@ -92,13 +92,16 @@ export const UserContextWrapper = (props: UserStoreContextProps) => {
         if (isUserDTO(data)) {
           setUser(data);
           setOrganizations(data.orgs);
-
           const orgId = getQueryParam("organization");
+          console.log(orgId);
+          console.log(data.orgs);
```
skeptrunedev, line 97:
```suggestion
```

skeptrunedev (changes requested):

chat/src/components/contexts/UserContext.tsx
aaryanpunia: fixed

skeptrunedev (approved):

PR #1792, opened 2024-07-05
feature: api version header

aaryanpunia:


skeptrunedev (changes requested):
The documentation is currently not being updated to include the X-API-VERSION header: 

image

Need to update the PR such that it's being included in the docs.

skeptrunedev:
These may or may not all be related to this PR, but we should still fix these Clippy warnings here: 

image

skeptrunedev (changes requested):
Fix clippy

skeptrunedev:
This header should be optional and not required. 

Also, this should not be present on all routes. I think for right now we should only be placing it on `search` and `recommendation` routes. 

image

server/src/middleware/api_version.rs
```diff
@@ -0,0 +1,119 @@
...
+{
+    type Response = ServiceResponse<B>;
+    type Error = Error;
+    type Future = LocalBoxFuture<'static, Result<Self::Response, Self::Error>>;
+
+    forward_ready!(service);
+
+    fn call(&self, mut req: ServiceRequest) -> Self::Future {
+        let path = req.path();
+        println!("{path}");
```
skeptrunedev, line 84:
```suggestion
```

skeptrunedev (changes requested):

skeptrunedev (changes requested):
image

I pushed a wip commit demonstrating the error. I think the issue is either ordering of middleware or just flat out not using the default value.

server/src/handlers/chunk_handler.rs
```diff
@@ -1101,6 +1103,7 @@ pub async fn search_chunks(
     pool: web::Data<Pool>,
     clickhouse_client: web::Data<clickhouse::Client>,
     dataset_org_plan_sub: DatasetAndOrgWithSubAndPlan,
+    api_version: APIVersion,
```
skeptrunedev, line 1106:
```suggestion
```

skeptrunedev (changes requested):
Also need to remove the newline where the print used to be

skeptrunedev (approved):

PR #1796, opened 2024-07-05
feature: all worker failures go to dead letter queue

aaryanpunia:

### DIFF
--- a/server/src/bin/ingestion-worker.rs
+++ b/server/src/bin/ingestion-worker.rs
@@ -1030,7 +1030,7 @@ pub async fn readd_error_to_queue(
     }
 
     if let IngestionMessage::BulkUpload(mut payload) = message {
-        let old_paylaod_message = serde_json::to_string(&payload).map_err(|_| {
+        let old_payload_message = serde_json::to_string(&payload).map_err(|_| {
             ServiceError::InternalServerError("Failed to reserialize input for retry".to_string())
         })?;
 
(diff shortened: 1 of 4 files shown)


server/src/bin/ingestion-worker.rs
```diff
@@ -1040,13 +1052,6 @@ pub async fn readd_error_to_queue(
             payload.attempt_number
         );
 
-        let _ = redis::cmd("LREM")
```
cdxker, line 1043: At this spot we should still LREM from processing, since it is going back into the ingestion queue

server/src/bin/ingestion-worker.rs
```diff
@@ -1019,6 +1019,18 @@ pub async fn readd_error_to_queue(
...
+                .await
+                .map_err(|err| ServiceError::BadRequest(err.to_string()))?;
+
+            redis::cmd("lpush")
+                .arg("dead_letters")
+                .arg(old_payload_message)
+                .query_async(&mut *redis_conn)
+                .await
+                .map_err(|err| ServiceError::BadRequest(err.to_string()))?;
+
```
cdxker, line 1033: We should also remove the item from the processing queue here, to keep the processing queue for only what is currently being processed

cdxker (changes requested):
Made comments for `ingestion-worker`, but should also be reflected for all other workersk

cdxker (approved):

PR #1797, opened 2024-07-05
enhancement: make count trigger run once for batch updates

aaryanpunia:


server/migrations/2024-07-05-221027_update_triggers/up.sql
```diff
@@ -0,0 +1,49 @@
...
+-- Function for INSERT operations
+CREATE OR REPLACE FUNCTION update_chunk_metadata_count_insert() 
+RETURNS TRIGGER AS $$
+BEGIN
+    INSERT INTO dataset_usage_counts (dataset_id, chunk_count)
+    SELECT dataset_id, COUNT(*)
+    FROM inserted
+    GROUP BY dataset_id
+    ON CONFLICT (dataset_id) DO UPDATE
+    SET chunk_count = dataset_usage_counts.chunk_count + EXCLUDED.chunk_count;
```
skeptrunedev, line 14: If it's ever the case that there are multiple dataset_id's then they will all get the same count.

skeptrunedev:
Closing in favor of the slightly cleaner implementation in #1798. 

That implementation uses only a single function for writes and variables for a bit better readability.

PR #1798, opened 2024-07-05
cleanup: update the count modification trigger function to run per statement instead of per row

skeptrunedev:


cdxker (changes requested):
We are NOT able to revert at these 2 locations with upsert_by_tracking_id. This reversion will delete the inserted chunk and all the groups associated with it. Add logic to avoid this reversion if upsert_by_tracking_id_being_used = true

https://github.com/devflowinc/trieve/pull/1798/files#diff-dae9db8e737e067728dd05badb0e26e12dd1a7d90132ff0649269abc915f65c6R664-R669 

https://github.com/devflowinc/trieve/pull/1798/files#diff-dae9db8e737e067728dd05badb0e26e12dd1a7d90132ff0649269abc915f65c6R664-R669

server/src/operators/chunk_operator.rs
```diff
@@ -1531,16 +1563,38 @@ pub async fn create_chunk_metadata(
...
+                    .collect::<Vec<String>>();
+
+                for group_tracking_id in group_tracking_ids {
+                    if !found_group_tracking_ids.contains(&group_tracking_id) {
+                        return Err(ServiceError::BadRequest(format!(
+                            "Group with tracking id {} does not exist",
+                            group_tracking_id
+                        )));
+                    }
+                }
```
cdxker, line 1592:
```suggestion
```
  whats the context behind this? If it stops at this point the full transaction will be reverted. We already check if the group_tracking_id exists in the handler here.    https://github.com/devflowinc/trieve/blob/adb27fea33e89553b6d2cb02ec7e6d86b1c54414/server/src/operators/chunk_operator.rs#L1547-L1597
  
 If it gets to this point and it doesn't exist, then they deleted it mid ingest which means we either ingested too slowly. Or a freak race condition occurred

cdxker (changes requested):

server/src/operators/chunk_operator.rs
skeptrunedev: Confusing function name, but this actually gets called from the handler. We already had a check for `group_id` but not `group_tracking_id`.

cdxker (approved):

PR #1814, opened 2024-07-08
cleanup: update docker-compose.yml

jooni22:
fix: change KEYCLOAK_BASE_URL to KEYCLOAK_HOST. 
Caused a problem launching the application on self-hosting


skeptrunedev (approved):
Thank you for the fix!

PR #1828, opened 2024-07-08
cleanup: prevent on conflict do update command cannot affect row a second time issues

skeptrunedev:


server/src/operators/chunk_operator.rs
```diff
@@ -531,15 +531,16 @@ pub async fn bulk_insert_chunk_metadata_query(
         .await
         .expect("Failed to get connection to db");
 
-    let chunk_metadata_to_insert: Vec<ChunkMetadataTable> = insertion_data
+    let chunk_metadatas_to_insert: Vec<ChunkMetadataTable> = insertion_data
         .clone()
         .iter()
         .map(|data| data.chunk_metadata.clone().into())
+        .unique_by(|chunk: &ChunkMetadataTable| (chunk.tracking_id.clone(), chunk.dataset_id))
```
cdxker, line 538: Make sure this only does it where TrackingId exists, if its None it might only insert 1 chunk

cdxker (changes requested):

server/src/operators/chunk_operator.rs
skeptrunedev: I think I fixed this now

cdxker (approved):

PR #1831, opened 2024-07-09
feature: download invoices

aaryanpunia:


skeptrunedev:
What are the CLI commands I need to run to test this? I don't have them on hand and it would help me to test efficiently.

skeptrunedev (changes requested):
With this implementation we will only ever create a single invoice for the subscription. 

We need to create additional invoices every time the subscription is charged on a recurring basis.

skeptrunedev:
Stripe dashboard has a feature which allows someone to advance time in order to simulate multiple charges to test this

skeptrunedev:
I changed the UI to put the invoices in the billing tab because it seems a bit cleaner: 

image

skeptrunedev (approved):

PR #1833, opened 2024-07-09
bugfix: KEYCLOAK_PROXY for docker hosting in VPS/VPC 

jooni22:
To fire up self-hosting with a domain on a VPS/VPC we need to set KEYCLOAK_PROXY=edge otherwise the application will not start correctly and the container server will report errors such as:
```
server | thread 'main' panicked at src/handlers/auth_handler.rs:148:6:
server | Failed to discover OpenID provider: response(308, [], "HTTP status code 308 Permanent Redirect at http://auth.yourdomain.co/realms/trieve/.well-known/openid-configuration")
server | note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
@skeptrunedev


skeptrunedev (approved):
Did a bit of cleanup to make this one commit and name the env variables in parity with what Keycloak accepts.

PR #1839, opened 2024-07-09
feat: analytics query table

drew-harris:


skeptrunedev:
Close as still in draft state

PR #1844, opened 2024-07-09
enhancement: use chunk/count method in chunk/search for get_total_pages

aaryanpunia:


skeptrunedev (dismissed):
This needs an update to run the count as a `futures::join` with the existing search such that they fire in parallel and not sequentially.

densumesh (approved):

densumesh (approved):

PR #1850, opened 2024-07-09
feature: reorganize analytics site

drew-harris:


drew-harris:
image

drew-harris:
image

densumesh:
image
when theres a generate off chunks this seems to overflow

image
make this table wider

Should make a new issue to make the dedicated query page to view a single query and its results

densumesh:
image

the bottom timestamps should correspond to the time range i.e. show the date rather than time stamp if looking at past week

Also if there is only 1 or 2 points make the graph have artifiically larger padding in terms of time so that the points aren't just at the side and are in the middle

densumesh:
image

Head queries should have a top bar showing what each row is as well

drew-harris:
Working now and bundle size is fixed. LGTM!

server/src/data/models.rs
```diff
@@ -3736,9 +3736,7 @@ pub enum SearchAnalytics {
         sort_order: Option<SortOrder>,
     },
     #[schema(title = "CountQueries")]
-    CountQueries {
-        filter: Option<SearchAnalyticsFilter>,
```
densumesh, line 3740: This should remain the same no?

densumesh (changes requested):

densumesh (approved):

densumesh (approved):

PR #1866, opened 2024-07-11
enhancement: update the local kubernetes guide and update the `helm` chart with latest changes for production

cdxker:


skeptrunedev (approved):
Niiice. Much easier to setup with minikube now.

PR #1867, opened 2024-07-11
feature: make bm25 sparse vector

cdxker:


densumesh (approved):

skeptrunedev (changes requested):
BM25 should be active by default for a dataset. We should use default values for the constants instead. 

If someone in an org for a dataset does nothing then they should still get BM25.

cdxker:
Just updated it to on the search UI

skeptrunedev (changes requested):
Add into `.env.search` for the `VITE_BM25_ACTIVE` env

.env.dashboard
```diff
@@ -3,3 +3,4 @@ VITE_SEARCH_UI_URL="http://localhost:5174"
 VITE_CHAT_UI_URL="http://localhost:5173"
 VITE_ANALYTICS_UI_URL="http://localhost:5176"
 VITE_SENTRY_DASHBOARD_DSN=""
+VITE_BM25_ACTIVE=false
```
skeptrunedev, line 6:
```suggestion
VITE_BM25_ACTIVE=true
```

.env.search
```diff
@@ -2,3 +2,4 @@
 VITE_API_HOST=http://localhost:8090/api
 VITE_SENTRY_SEARCH_DSN=""
 VITE_DASHBOARD_URL=http://localhost:5173
+VITE_BM25_ACTIVE=false
```
skeptrunedev, line 5:
```suggestion
VITE_BM25_ACTIVE=true
```

skeptrunedev (changes requested):

skeptrunedev (changes requested):
Need to have BM25 in update

skeptrunedev (approved):

PR #1874, opened 2024-07-11
feature: max limit in server config

aaryanpunia:


skeptrunedev (changes requested):
Not in the issue, but there should be a tooltip added next to max limit explaining what it does since this isn't really intuitive. 

image

Directly from the issue: 

>  pass-through for get_total_pages is implemented such that the limits are enforced as defaults instead of blocking the request


This is pretty critical to get in this PR.

skeptrunedev (approved):

PR #1878, opened 2024-07-11
feature: refactor types for API V2 release

densumesh:


skeptrunedev (changes requested):
Redoc should have V2 as the default response type

skeptrunedev (changes requested):
- get_recommended_groups
- get_recommended_chunks

Should also be changed to have the same response type as the search routes do now

skeptrunedev:
Before we merge this we should also modify `tag_set` on these reponse types to be an array

skeptrunedev (changes requested):
image

`search_over_groups` still needs a revision

skeptrunedev:
I am happy with where this PR is at now afaik. Want to go through it one more time reviewing all the search and recommendation routes then I think it's good to merge. Extremely excited!!

skeptrunedev:
### Verify we default to old response and that new response is satisfactory for: 

- [ ] autocomplete
- [ ] search
- [ ] get recommended chunks
- [ ] search over groups
- [ ] get recommended groups
- [ ] search within group

skeptrunedev:
> ### Verify we default to old response and that new response is satisfactory for:
> * [x]  autocomplete
> * [x]  search
> * [x]  get recommended chunks
> * [x]  search over groups
> * [x]  get recommended groups
> * [x]  search within group

Finished check and believe this is good to go. Will merge a bit later tonight after back from errands!!

skeptrunedev (approved):

PR #1880, opened 2024-07-11
feature: reranker flag

aaryanpunia:


skeptrunedev (changes requested):
If using the re-ranker the scores should be replaced with that of the `cross_encoder`

skeptrunedev (changes requested):
`score_threshold` should apply at the post-reranker level instead of in-index when `use_reranker: true` is set

skeptrunedev (approved):

PR #1887, opened 2024-07-12
styling: sensible timestamp labels for analytics graphs

drew-harris:


densumesh:
image
I think the ticks for per hours should be the last 60 minutes

densumesh:
image
should be by day as well when you look at the past week

densumesh (approved):

PR #1891, opened 2024-07-12
bugfix: dont error if one item in batch is empty

densumesh:


skeptrunedev (changes requested):
Let's error in the handler if too large of chunks are getting ingested **or** there are empty chunks. 

And not modify the ingestion-worker.

PR #1894, opened 2024-07-12
fix: dedup invoice by hosted_url

aaryanpunia:


skeptrunedev (changes requested):
Are you sure that you are still able to subscribe to a plan on this branch? 

For some reason, it's just not subscribing me at all now after a successful stripe purchase. Is the error message return the issue?

skeptrunedev (approved):

PR #1906, opened 2024-07-13
bugfix: delete the group_id within Qdrant for relevant chunks when a group is deleted

skeptrunedev:
- **bugfix: delete the group_id within Qdrant for relevant chunks when a group is deleted**
- **cleanup: properly keep track of dataset on links in navbar and org-groups and groups page**


cdxker:
Wouldn't this be to slow to finish within a single request?

cdxker (approved):

PR #1921, opened 2024-07-15
cleanup: get server config from db in all workers as to not pollute the message

aaryanpunia:


server/src/bin/delete-worker.rs
```diff
@@ -216,13 +219,32 @@ async fn delete_worker(
...
+                delete_worker_message,
+                dataset.err().unwrap(),
+                redis_pool.clone(),
+            )
+            .await;
+            return;
+        }
+
+        let server_config =
+            ServerDatasetConfiguration::from_json(dataset.unwrap().server_configuration);
```
skeptrunedev, line 239: Instead of the `unwrap` in these we should use a `match` and handle the err correctly w/in the match so we can guarantee not crash the worker server processes.

skeptrunedev (changes requested):
The `unwrap`'s here are generally a bad idea and we shouldn't do that. Using a `match` is a much better idea.

skeptrunedev (approved):

PR #1922, opened 2024-07-15
bugfix: navigate to group list if dataset changes and group doesn't exist

aaryanpunia:


frontends/search/src/components/GroupPage.tsx
```diff
@@ -154,6 +154,7 @@ export const GroupPage = (props: GroupPageProps) => {
       let group_id: string | null = null;
       const currentDataset = $dataset?.();
       if (!currentDataset) return;
+      console.log("effect ran");
```
skeptrunedev, line 157:
```suggestion
```

frontends/search/src/components/GroupPage.tsx
```diff
@@ -173,6 +174,8 @@ export const GroupPage = (props: GroupPageProps) => {
               setTotalPages(groupBookmarks.total_pages);
               setChunkMetadatas(groupBookmarks.chunks);
             });
+          } else if (response.status == 404) {
+            navigate("/group");
```
skeptrunedev, line 178: I don't think we should forcibly navigate because that's confusing to the easier. Imo, we should just clear out the data.

skeptrunedev (changes requested):
Going to ship a removal of the redirect real quick and re-request my review. Will be a simple amend.

skeptrunedev (approved):

PR #1923, opened 2024-07-15
feature: tracking id in search ui

aaryanpunia:


skeptrunedev (approved):
Crazy that we already had `trackingId` as a signal for `EditChunkPageForm`. 

Excited this is finally in the search playground :sunglasses: :smile: .

PR #1924, opened 2024-07-15
feature: scope api keys to certain routes

densumesh:


skeptrunedev (changes requested):
I think we should add this to the dashboard in the same PR? Seems like that would be the way most people would want to use this.

skeptrunedev:
> I think we should add this to the dashboard in the same PR? Seems like that would be the way most people would want to use this.

I take this back bc https://github.com/devflowinc/trieve/issues/1925 was made. Will do this in a diff PR

skeptrunedev (approved):

PR #1927, opened 2024-07-16
feature: reindex worker for bm25 and process to queue bm25 reinndexing

cdxker:
This PR has to new scripts 
`reindex-worker.rs` This receives a Migration point message.

```rs
#[derive(Debug, Clone, Deserialize, Serialize)]
pub enum MigrationMode {
    BM25 { average_len: f32, k: f32, b: f32 },
}

#[derive(Debug, Clone, Deserialize, Serialize)]
pub struct MigratePointMessage {
    pub qdrant_point_ids: Vec<uuid::Uuid>,
    pub to_collection: String,
    pub from_collection: String,
    pub mode: MigrationMode,
}
```
`reindex-worker.rs` simply just moves a set of points from one collection to another, and optionally changing the vector. 

`queue-bm25-migration.rs` 

Which simple scrolls through all collections defined by `VECTOR_SIZES` env that already exists. And submit's bm25 migrations from `vectors_<vector_size>` to `vectors_<vector_size>_bm25`. 

In the future cleanup will need to be done to remove the point in the other collection. For safety this does not delete the original points and if a failure happens it is recorded in a dead_letter queue. No retries


server/src/bin/delete-worker.rs
```diff
@@ -393,7 +393,7 @@ pub async fn readd_error_to_queue(
             .map_err(|err| ServiceError::BadRequest(err.to_string()))?;
 
         redis::cmd("lpush")
-            .arg("dead_letters")
+            .arg("dead_letters_delete")
```
skeptrunedev, line 396: Oh damn, this is a really nice fix. Didn't catch this on the original PR

skeptrunedev (commented):
Will this create the secondary BM25 collections automatically?

cdxker:
> Will this create the secondary BM25 collections automatically?

No, new collections will have to be explicitly created. Didn't included that in this pr since some people will want 2x replication, different number of shards, etc. Better to just be generic

skeptrunedev (approved):
LGTM 

Kind of a big PR and I couldn't go through it super in depth, but it seems to work. 

Also, the only real risk is Qdrant OOM'ing during the migration which we can just monitor. Further, replicated setups will stay up even given the OOM'ing if it happens.

PR #1931, opened 2024-07-16
feature: add route to organization for updating the ServerConfiguration of all datasets

skeptrunedev:
Also ships some nice doc updates for the dataset configs so it's easier to understand what you can set for the dataset config. 

Since removing the `ClientConfiguration` the `server` prefix was no longer needed, so also dropped that throughout the code base. Will figure out a migration for the table and dataset create/update routes later.

### DIFF
--- a/server/src/operators/organization_operator.rs
+++ b/server/src/operators/organization_operator.rs
@@ -1,18 +1,17 @@
 use crate::{
     data::models::{
-        Dataset, Organization, OrganizationUsageCount, OrganizationWithSubAndPlan, Pool, RedisPool,
-        ServerDatasetConfiguration, SlimUser, StripePlan, StripeSubscription, User,
-        UserOrganization,
+        Dataset, DatasetConfiguration, Organization, OrganizationUsageCount,
+        OrganizationWithSubAndPlan, Pool, RedisPool, SlimUser, StripePlan, StripeSubscription,
+        User, UserOrganization,
     },
     errors::ServiceError,
     operators::dataset_operator::soft_delete_dataset_by_id_query,
     randutil,
 };
 use actix_web::{web, HttpRequest};
-use diesel::prelude::*;
 use diesel::{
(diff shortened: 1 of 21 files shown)


skeptrunedev:
I tested this on 25k datasets locally and the latency was consistently in the 20s range.

server/src/operators/organization_operator.rs
```diff
@@ -622,3 +622,47 @@ pub async fn delete_actual_organization_query(
...
+            .execute(&mut conn)
+            .await
+            .map_err(|e| {
+                log::error!(
+                    "Error updating dataset in update_all_org_dataset_server_configs: {:?}",
+                    e
+                );
+                ServiceError::BadRequest("Error updating dataset".to_string())
+            })?;
+    }
```
skeptrunedev, line 665: Want to draw attention to this section. 

I chose to do multiple small transactions instead of one big transaction because I didn't want to ever lock the whole table. 

The downside to my approach is there it incurs overhead from so many commits for orgs with lots of datasets, but I still think it's worth avoiding the big lock. 

Other consideration is that this in theory possible with one dataset call which I want to try before this gets merged. Will be attempting here shortly. 

I think postgres supports the concat operator on a update even tho diesel does not.

server/src/operators/organization_operator.rs
```diff
@@ -622,3 +621,29 @@ pub async fn delete_actual_organization_query(
...
+        .map_err(|e| {
+            log::error!(
+                "Error updating datasets in update_all_org_dataset_server_configs: {:?}",
+                e
+            );
+            ServiceError::BadRequest("Error updating datasets".to_string())
+        })?;
+
+    Ok(())
+}
```
skeptrunedev, line 649: After making this change we are down from 20s to 200ms: 

image

cdxker: What were you doing before? The commit got force pushed over

cdxker (approved):

PR #1934, opened 2024-07-16
cleanup: do not use quote and negated terms by default + documentation: add docs for TR-Dataset to specify a tracking_id can be used"

skeptrunedev:
- **cleanup: do not use quote negated terms by default**
- **documentation: add docs to the TR-Dataset headers to note that we accept tracking_id's as valid values**

### DIFF
--- a/server/src/handlers/organization_handler.rs
+++ b/server/src/handlers/organization_handler.rs
@@ -328,6 +328,35 @@ pub async fn remove_user_from_org(
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone, ToSchema)]
+#[schema(example = json!({
+    "organization_id": "00000000-0000-0000-0000-000000000000",
+    "server_configuration": {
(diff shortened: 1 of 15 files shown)


server/src/handlers/organization_handler.rs
```diff
@@ -328,6 +328,35 @@ pub async fn remove_user_from_org(
...
+        "FREQUENCY_PENALTY": 0.0,
+        "TEMPERATURE": 0.5,
+        "PRESENCE_PENALTY": 0.0,
+        "STOP_TOKENS": ["\n\n", "\n"],
+        "INDEXED_ONLY": false,
+        "LOCKED": false,
+        "SYSTEM_PROMPT": "You are a helpful assistant",
+        "MAX_LIMIT": 10000
+    }
+}))]
```
skeptrunedev, line 359: Brings this route inline with the dataset resource routes which already had this doc example. Hard to figure out what can be specified without it.

cdxker (approved):

PR #1940, opened 2024-07-16
bugfix: dont remove `\`

densumesh:

### DIFF
--- a/server/src/operators/organization_operator.rs
+++ b/server/src/operators/organization_operator.rs
@@ -629,7 +629,7 @@ pub async fn update_all_org_dataset_configs_query(
 ) -> Result<(), ServiceError> {
     let concat_configs_raw_query = sql_query(format!(
         "UPDATE datasets SET server_configuration = server_configuration || '{}' WHERE organization_id = '{}';",
-        new_config.to_string().replace('\'', "''").replace('\\', ""), org_id
+        new_config.to_string().replace('\'', "''"), org_id
     ));
 
     let mut conn = pool.get().await.unwrap();


server/src/operators/organization_operator.rs
```diff
@@ -629,7 +629,7 @@ pub async fn update_all_org_dataset_configs_query(
 ) -> Result<(), ServiceError> {
     let concat_configs_raw_query = sql_query(format!(
         "UPDATE datasets SET server_configuration = server_configuration || '{}' WHERE organization_id = '{}';",
-        new_config.to_string().replace('\'', "''").replace('\\', ""), org_id
+        new_config.to_string().replace('\'', "''"), org_id
```
skeptrunedev, line 632:
```suggestion
        new_config.to_string().replace('\'', "''"), org_id
```

skeptrunedev: Shouldn't there only be one single quote in the second param spot?

skeptrunedev (changes requested):

server/src/operators/organization_operator.rs
densumesh: No to escape the ' in Postgres you double quote

densumesh: https://stackoverflow.com/questions/12316953/insert-text-with-single-quotes-in-postgresql

skeptrunedev: Damn, alright

skeptrunedev (approved):

PR #1947, opened 2024-07-17
feat: hide qdrant point ids from responses

drew-harris:


skeptrunedev (approved):
This is cool. Tested locally and it works though `cargo watch` won't catch a serde comment to recompile by default. 

Extremely clean solution.

PR #1949, opened 2024-07-17
observerability: add timestamp logs back to workers

cdxker:
I thought excluding times was a good idea since we have a log sync solution with k9s, but k9s is not a good solution for logs with timestamps.

Adding back explicit timestamps for now


skeptrunedev (changes requested):
amend'ing cargo fmt fix

skeptrunedev (approved):

PR #1954, opened 2024-07-17
fix: remove replace in AfMessage

aaryanpunia:


skeptrunedev (approved):

skeptrunedev:
Amend'ed to remove a console log, but otherwise looks good! Kind of cool that the markdown formatter roughly knows how to display these. 

image

PR #1955, opened 2024-07-17
feature: nav to group page from group search

aaryanpunia:


skeptrunedev (approved):
In a force push amend I added styling so it's in the middle of the collapsible section and also added the `?dataset=` bit to the link to guarantee that it wouldn't switch datasets on click.

PR #1957, opened 2024-07-17
feature: grpc with tei

aaryanpunia:


server/src/operators/model_operator.rs
```diff
@@ -1030,3 +1058,494 @@ pub fn term_frequency(
...
+
+pub mod tei {
+    tonic::include_proto!("tei.v1");
+}
+
+async fn create_batch_embedding_call(
+    messages: Vec<String>,
+    channel_to_use: Option<Channel>,
+) -> Result<Vec<Vec<f32>>, ServiceError> {
+    let grpc_origin = get_env!("EMBEDDING_SERVER_GRPC_ORIGIN", "Grpc origin should be set");
```
cdxker, line 1070: How will this work with all the different types of embedding servers?

server/src/operators/model_operator.rs
```diff
@@ -1030,3 +1058,494 @@ pub fn term_frequency(
...
+        .iter()
+        .enumerate()
+        .filter_map(|(index, boost)| {
+            boost
+                .clone()
+                .map(|distance_phrase| (index, distance_phrase.phrase))
+        })
+        .unzip();
+
+    let grpc_origin = get_env!("EMBEDDING_SERVER_GRPC_ORIGIN", "Grpc origin should be set");
```
cdxker, line 1229: Remove `get_env!` and instead use `std::env::var()` since this is an optional parameter

server/src/operators/model_operator.rs
```diff
@@ -1030,3 +1058,494 @@ pub fn term_frequency(
...
+    }
+
+    Ok(combined_vecs)
+}
+
+pub async fn get_sparse_vector_grpc(
+    message: String,
+    _embed_type: &str,
+) -> Result<Vec<(u32, f32)>, ServiceError> {
+    let grpc_origin = get_env!("SPARSE_SERVER_GRPC_ORIGIN", "Grpc origin should be set");
```
cdxker, line 1266: Same as above, use std::env::var here instead

server/src/operators/model_operator.rs
```diff
@@ -1030,3 +1058,494 @@ pub fn term_frequency(
...
+                _ => Err(ServiceError::BadRequest("Metadata not found".to_string())),
+            }?;
+
+            Ok(convert_html_to_text(
+                &(chunk.chunk_html.unwrap_or_default()),
+            ))
+        })
+        .collect::<Result<Vec<String>, ServiceError>>()?;
+
+    let grpc_origin = get_env!("RERANKER_SERVER_GRPC_ORIGIN", "Grpc origin should be set");
```
cdxker, line 1345: .

server/src/operators/model_operator.rs
```diff
@@ -1030,3 +1058,494 @@ pub fn term_frequency(
...
+        .map(|(x, _)| x)
+        .collect::<Vec<String>>();
+    let thirty_content_groups = contents.chunks(30);
+    let filtered_boosts_with_index = messages
+        .into_iter()
+        .enumerate()
+        .filter_map(|(i, (_, y))| y.map(|boost_phrase| (i, boost_phrase)))
+        .collect::<Vec<(usize, BoostPhrase)>>();
+    let filtered_boosts_with_index_groups = filtered_boosts_with_index.chunks(30);
+    let grpc_origin = get_env!("SPARSE_SERVER_GRPC_ORIGIN", "Grpc origin should be set");
```
cdxker, line 1402: .

server/src/operators/model_operator.rs
```diff
@@ -1030,3 +1058,494 @@ pub fn term_frequency(
...
+    }
+
+    Ok(content_vectors_sorted)
+}
+
+pub async fn get_batch_sparse_vectors_grpc(
+    messages: Vec<String>,
+    channel_to_use: Option<Channel>,
+) -> Result<Vec<Vec<(u32, f32)>>, ServiceError> {
+    let grpc_origin = get_env!("SPARSE_SERVER_GRPC_ORIGIN", "Grpc origin should be set");
```
cdxker, line 1497: .

cdxker (changes requested):
We may need to have a different strategy with how we will point to the other embedding servers. Maybe we just use the pre-existing env's `origin` env's and just swap if use_grpc is active?

cdxker:
The testing needed for this is:

1) Start up the server without making any additional env changes and for the server to use http for embedding requests and ingestion messages.
2) Then add additional env's to have it point to grpc routes then rerun the same queries.

aaryanpunia:
Embedding calls are not being done in batches, getting batch size too large

skeptrunedev:
We **should** merge this PR, but we aren't going to right now because it seems complicated with helm for the services and the migration is intensive. There are other things higher in prio.

PR #1966, opened 2024-07-18
fix: undefined instead of empty arrays while generating api keys in d…

aaryanpunia:


skeptrunedev (approved):

skeptrunedev (approved):
Was going to add a tooltip to describe what happens on hover, but realized we already have one!

PR #1974, opened 2024-07-18
feature: utilize hdbscan for clustering

aaryanpunia:


docker/clustering-script/get_clusters.py
```diff
@@ -6,9 +6,12 @@
 import clickhouse_connect.driver
 import clickhouse_connect.driver.client
 import numpy as np
-from sklearn.cluster import KMeans
+from sklearn.cluster import HDBSCAN
 from scipy.spatial.distance import cosine
-anthropic_client = anthropic.Anthropic()
+import dotenv
+
+dotenv.load_dotenv()
```
densumesh, line 13:
```suggestion

```
Need to remove dotenv so that the env vars work correctly in docker

docker/clustering-script/get_clusters.py
```diff
@@ -6,9 +6,12 @@
...
+anthropic_client = anthropic.Anthropic(api_key=os.getenv("LLM_API_KEY"))
```
densumesh, line 14: Either need to modify the Dockerfile to use this new env var or just use `ANTHROPIC_API_KEY` which it automatically pulls.

densumesh (changes requested):

docker/clustering-script/get_clusters.py
cdxker: This is fine

cdxker (changes requested):

densumesh (approved):

cdxker (approved):

PR #1985, opened 2024-07-22
feature: add batch group create

skeptrunedev:


server/src/handlers/group_handler.rs
```diff
@@ -93,35 +134,77 @@ pub struct CreateChunkGroupReqPayload {
...
+    let created_groups = upsert_results?
+        .into_iter()
+        .chain(non_upsert_results?.into_iter())
+        .collect::<Vec<ChunkGroup>>();
+
+    if created_groups.len() == 1 {
+        Ok(HttpResponse::Ok().json(created_groups[0].clone()))
+    } else {
+        Ok(HttpResponse::Ok().json(created_groups))
+    }
```
skeptrunedev, line 207: I would rather always return an array, but I wrote it this way for the sake of preserving existing deployments which rely on a single group returned.

densumesh, line 207: Prob should keep the response type consistent/document that it's different if bulk

densumesh (changes requested):

server/src/handlers/group_handler.rs
skeptrunedev: Docs are actually updated

server/src/handlers/group_handler.rs
```diff
@@ -71,18 +82,48 @@ pub struct CreateChunkGroupReqPayload {
...
     path = "/chunk_group",
     context_path = "/api",
     tag = "Chunk Group",
-    request_body(content = CreateChunkGroupReqPayload, description = "JSON request payload to cretea a chunkGroup", content_type = "application/json"),
+    request_body(content = CreateChunkGroupReqPayloadEnum, description = "JSON request payload to cretea a chunk_group(s)", content_type = "application/json"),
     responses(
-        (status = 200, description = "Returns the created chunkGroup", body = ChunkGroup),
-        (status = 400, description = "Service error relating to creating the chunkGroup", body = ErrorResponseBody),
+        (status = 200, description = "Returns the created chunk_group if a single chunK_group was specified", body = ChunkGroup),
+        (status = 200, description = "Returns the created chunk_groups if a batch of chunk_groups was specified", body = Vec<ChunkGroup>),
```
skeptrunedev, line 125: @densumesh I added docs for both resp types

server/src/handlers/group_handler.rs
skeptrunedev: https://github.com/devflowinc/trieve/pull/1985#discussion_r1686010360

cdxker (approved):

densumesh (approved):

PR #1987, opened 2024-07-22
feature: File upload button in new dataset (dashboard ui)

aaryanpunia:


skeptrunedev (changes requested):
1. Put the `FileUploadModal` in shared components
~~2. This needs to pass CI~~

skeptrunedev (changes requested):
Last thing before I actually test functionality is to collapse this down into one commit please.

skeptrunedev (changes requested):
This should be refactored to go with a simpler implementation where we just link out to the upload file page in the search playground for the appropriate dataset. 

This really just gets us started in a bad direction where we start trying to replicate functionality in dashboard which already exists in search and that's painful. 

Much much easier to just link out to the search playground here. I'm going to quickly implement that change.

skeptrunedev (approved):
Nice! Sorry again about the pain on doing this twice :( .

PR #2002, opened 2024-07-23
feature: remove compression

densumesh:


skeptrunedev (approved):

skeptrunedev:
Tested with CURL that this actually does compress for search and recommend routes. Also ran integration test on streaming for both generate_from_chunks and create_message. Verified streaming is working for both of those routes as well.

PR #2014, opened 2024-07-24
cleanup: prevent user from creating multiple groups with the same tracking_id

skeptrunedev:


cdxker (changes requested):
Also make sure the tracking if isn't an empty string

skeptrunedev:
> Also make sure the tracking if isn't an empty string

Yeah, I missed that. Added now :checkered_flag:

cdxker (approved):

PR #2016, opened 2024-07-24
feature: new highlight system

aaryanpunia:


skeptrunedev (changes requested):
It's not quite working like I would expect it to:

image

Something seems off regarding the implementation not replacing each token which it should.

skeptrunedev (approved):

PR #2019, opened 2024-07-25
feature: update filters to have match_any and match_all

densumesh:


skeptrunedev (approved):
Nice! Tested with search playground and it looks good.

PR #2023, opened 2024-07-26
feature: validate dataset configs

densumesh:


skeptrunedev:
It would be nice to get this, but there are Vacuum limitations which would take a while to overcome and the cleanup isn't worth breaking the Redoc CI action.

PR #2025, opened 2024-07-26
fix: replace chunk_by with for loop

aaryanpunia:


skeptrunedev (approved):
Less Rust'y, but should work with Rust 1.75

PR #2027, opened 2024-07-28
docs: update api endpoint of query suggestions

SaraVieira:
Api docs are currently wrong [here](https://docs.trieve.ai/api-reference/chunk/generate-suggested-queries), and it seems the open api json is also not updated [here](https://api.trieve.ai/api-docs/openapi.json) and this fixes the second part and the open API endpoint

cc @densumesh @skeptrunedev


skeptrunedev (approved):
Sad this was broken for so long :(. 

But happy it's fixed now!

Issue #2031, opened 2024-07-29
cleanup: use non-prefix expanded index for quote and negated word filters 


skeptrunedev:
### Description

Since switching to the prefix expanded index for autocomplete, the experience for quote and negated words has been off. It needs to use a non prefix expanded index with Qdrant. 

### Target(s)

`server`

### Requirement to close

PR to use a non prefix expanded index for quote negated word filters 

### Community channels

Matrix is preferred. Reach out on discord or Matrix for further assistance. 

- [discord](https://discord.gg/CuJVfgZf54)
- [matrix](https://matrix.to/#/#trieve-general:matrix.zerodao.gg)

densumesh:
Can't create multiple indexes on one field

skeptrunedev:
Not currently possible to solve this with Qdrant

PR #2037, opened 2024-07-29
feature: only create indexes if env flag is set

densumesh:


skeptrunedev (commented):
Will this still use the prefix tokenizer as the payload filter for autocomplete?

densumesh:
> Will this still use the prefix tokenizer as the payload filter for autocomplete?

yes. cc https://github.com/devflowinc/trieve/issues/2031#issuecomment-2256424917

skeptrunedev:
Given we already have `CREATE_QDRANT_COLLECTIONS`, I don't think this adds much value and am going to close. Imo, this really just adds complication.

PR #2039, opened 2024-07-29
enhancement: loading spinner for login

aaryanpunia:
Tested for - 
1. User has logged out and tries to open the page: redirects back to keycloak (no changes necessary)
2. User is logged in and opens page: spinner (enhancement)
3. User does not have required role: You do not have access screen (no change)

### DIFF
--- a/frontends/dashboard/src/contexts/UserContext.tsx
+++ b/frontends/dashboard/src/contexts/UserContext.tsx
@@ -25,6 +25,7 @@ export interface UserStore {
   setSelectedOrganizationId: (id: string | null) => void;
   login: () => void;
   logout: () => void;
+  loading: Accessor<boolean> | null;
 }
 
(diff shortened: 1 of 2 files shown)


aaryanpunia:
https://github.com/user-attachments/assets/94789791-52fd-4603-a8c2-227137372588

skeptrunedev (changes requested):
I may have been a bit too unclear on the issue. 

The current problem is that the redirect to keycloak takes too long and the toast shows.

skeptrunedev:
> I may have been a bit too unclear on the issue.
> 
> The current problem is that the redirect to keycloak takes too long and the toast shows.

In context this fixes a different issue where when signed in it flashes an unauthorized while things load in.

skeptrunedev (approved):

PR #2051, opened 2024-07-30
cleanup: highlight refactor

aaryanpunia:


skeptrunedev (changes requested):
nested highlights came back

skeptrunedev (changes requested):
It seems like there was a regression on the previous error case: 

image

skeptrunedev (approved):

PR #2066, opened 2024-07-31
wip: fix highlights

aaryanpunia:
very difficult to say what to do next,

the reason we get a lot of highlights for trieve docs, is that the search engine returns results. So it's actually functioning correctly. This pr mostly tries to selectively truncate highlight windows, some work has to be done to correctly decide the min max part of calculating merged_results. The other thing is whether we want to actually truncate the array based on max windows after doing all our merges, because it seems like results were better before we did that ?

### DIFF
--- a/server/src/operators/chunk_operator.rs
+++ b/server/src/operators/chunk_operator.rs
@@ -1,4 +1,4 @@
-use std::cmp::Ordering;
+use std::cmp::{max, min_by_key, Ordering};
 use std::collections::{HashMap, HashSet};
 
 use crate::data::models::{
@@ -1390,6 +1390,7 @@ pub fn get_highlights_with_exact_match(
             false => Some(chunk.join(" ")),
         })
         .filter(|x| !x.is_empty())
+        .map(|p| p.to_lowercase())
         .collect_vec();
 
     let content = convert_html_to_text(&(input.chunk_html.clone().unwrap_or_default()));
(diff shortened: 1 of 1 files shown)


server/src/operators/chunk_operator.rs
```diff
@@ -1548,21 +1555,60 @@ pub fn get_highlights_with_exact_match(
...
+                            prev_w_start,
+                            run.1.saturating_sub(1).min(prev_w_end),
+                            combined_phrases.clone().into_iter().unique().collect_vec(),
+                        ))
+                    }
+                    prev_w_start = run.3.max(prev_w_end).min(run.1);
+                    prev_w_end = run.4;
+                    combined_phrases = vec![run.2.to_string()];
+                }
+                prev_was_exact = is_exact;
```
aaryanpunia, line 1599: trying to truncate windows selctively here, need to correctly decide what `prev_w_start` should be and what the `else` branch should be pushing to get the results we want

server/src/operators/chunk_operator.rs
```diff
@@ -1438,22 +1439,28 @@ pub fn get_highlights_with_exact_match(
...
         .enumerate()
-        .filter_map(
-            |(i, split)| match query_parts_split_by_stop_words.contains(split) {
-                true => Some((i, split.clone())),
+        .filter_map(|(i, split)| {
+            match query_parts_split_by_stop_words.contains(&split.to_lowercase()) {
+                true => {
+                    query_parts_used.insert(split.clone());
+                    Some((i, split.clone()))
+                },
```
aaryanpunia, line 1452: This solves an actual bug, this should remain mostly because we were incorrectly not lowercasing our query and splits when comparing previously

server/src/operators/chunk_operator.rs
```diff
@@ -1438,22 +1439,28 @@ pub fn get_highlights_with_exact_match(
...
+            }
+        })
         .collect_vec();
 
     let exact_matches = matched_idxs.clone();
 
     let results: Vec<usize> = query_parts_split_by_stop_words
         .clone()
         .iter()
+        .filter(|p| !query_parts_used.contains(&p.to_string()))
```
aaryanpunia, line 1463: whether or not we should do this is debatable, but i think this let's us avoid the search engine giving us results that may overlap with exact matches

PR #2069, opened 2024-07-31
styling: add icons to indicate clickable rows in analytics

drew-harris:


drew-harris:
image

Didn't spend too much time on it since we are doing the big table page anyway

skeptrunedev (approved):
Did a small amend to move expand right and add a hover state 

Agree with not spending much time given the tables are coming 

image

PR #2072, opened 2024-07-31
feature: validate dataset configs

densumesh:


skeptrunedev (approved):
LET'S GOOO! 

One of those latent issues with the docs that sucked up to now.

PR #2074, opened 2024-07-31
cleanup: remove duplicate strings from API reference docs

vid277:


skeptrunedev (approved):
Amend'ed to add hover state to expand icon + move expand icon right for suggested queries 
image

PR #2078, opened 2024-08-01
feature: keycloak and clickhouse persist

cdxker:
Keycloak now persists:
- [X] Across helm install / helm uninstall (by using postgres)
- [X] When postgresql volume is increased

Clickhouse now persists:
- [X] On volume increase
- [ ] ~~Across helm install / uninstall~~ (Was not able to achieve)

Punting on Clickhouse persisting across helm installs, thinking to add deletion protection the ebs volume on AWS.


cdxker:
Keycloak now persists:
- [X] Across helm install / helm uninstall (by using postgres)
- [X] When postgresql volume is increased

Clickhouse now persists:
- [X] On volume increase
- [ ] ~~Across helm install / uninstall~~ (Was not able to achieve)

PR #2080, opened 2024-08-01
feature: add aryn file chunking

densumesh:


skeptrunedev (changes requested):
Need to add new env to `.env.server` and `docker-compose.yml`. 

Also can you add a flag to the frontend? It would be nice to be able to do a comparison video.

skeptrunedev:
Going to close for now. It seems like the optimal way to do this is outside of Trieve and not within. During testing it was not clear that layout model based chunking beat out splitting.

PR #2088, opened 2024-08-01
enhancement: new highlight strategy

aaryanpunia:


skeptrunedev (changes requested):
- change semantics here to "rate"
- add ability for user to include a note
- after clicking the "rate query" button show a FullScreenModal where they can add the score and note (1-10 for score)

skeptrunedev (changes requested):
Problem on the `Unexpected server response: 503` query where `503` doesn't get highlighted. Will review tmrw.

skeptrunedev:
Closing in favor of #2110

PR #2094, opened 2024-08-02
feature: delete files and groups with clear route

densumesh:
This PR modifies clear to remove all data including files and groups on clear then additionally not delete anything which is added to the dataset after the clear operation occurs.

### DIFF
--- a/server/src/operators/dataset_operator.rs
+++ b/server/src/operators/dataset_operator.rs
@@ -184,6 +184,7 @@ pub async fn get_dataset_and_organization_from_dataset_id_query(
 pub struct DeleteMessage {
     pub dataset_id: uuid::Uuid,
     pub attempt_number: usize,
+    pub deleted_at: chrono::NaiveDateTime,
     pub empty_dataset: bool,
 }
 
@@ -224,6 +225,7 @@ pub async fn soft_delete_dataset_by_id_query(
     let message = DeleteMessage {
         dataset_id: id,
         attempt_number: 0,
+        deleted_at: chrono::Utc::now().naive_utc(),
         empty_dataset: false,
     };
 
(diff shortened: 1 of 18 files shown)


server/src/operators/dataset_operator.rs
```diff
@@ -274,18 +277,67 @@ pub async fn clear_dataset_by_dataset_id_query(
...
+            .filter(files_column::dataset_id.eq(id))
+            .filter(files_column::created_at.lt(deleted_at)),
+    )
+    .execute(&mut conn)
+    .await
+    .map_err(|err| {
+        log::error!("Could not delete groups: {}", err);
+        ServiceError::BadRequest("Could not delete groups".to_string())
+    })?;
+
```
skeptrunedev, line 340: I worry about this for datasets which have 10k+ groups, but idk. I think it's fine for now, but something to think about if we see usage on these resources start to spike.

server/src/operators/dataset_operator.rs
```diff
@@ -294,6 +346,7 @@ pub async fn delete_chunks_in_dataset(
             chunk_metadata_columns::chunk_metadata
                 .filter(chunk_metadata_columns::dataset_id.eq(id))
                 .filter(chunk_metadata_columns::id.gt(last_offset_id))
+                .filter(chunk_metadata_columns::created_at.lt(deleted_at))
```
skeptrunedev, line 349: What's the point of this?

skeptrunedev: Actually nice touch. The point of this is to make sure that we don't delete data which has been added post the clear call. This keeps that data safe if you are writing and clearing at the same time.

skeptrunedev:
Added a commit doing some icon things and cleaning up function name and logs relating to this

skeptrunedev (approved):

PR #2097, opened 2024-08-02
fix: always clip messages

aaryanpunia:


skeptrunedev (approved):
Will hopefully prevent no-op writes from occurring in the ingestion worker.

PR #2105, opened 2024-08-02
feature: upload multiple files or an entire folder on search playground

vid277:


skeptrunedev (changes requested):
image

While group count is changing, all chunks are being placed into the same group and no new group is being created for each file.

densumesh (approved):

densumesh (approved):

PR #2109, opened 2024-08-02
feature: add multi query with weight to semantic search

densumesh:

### DIFF
--- a/frontends/search/src/components/ResultsPage.tsx
+++ b/frontends/search/src/components/ResultsPage.tsx
@@ -212,9 +212,16 @@ const ResultsPage = (props: ResultsPageProps) => {
           : (sort_by = undefined);
       }
 
+      const query =
+        props.search.debounced.multiQueries.length > 0
+          ? props.search.debounced.multiQueries
+              .map((q) => [q.query, q.weight])
+              .filter((q) => q[0] != "")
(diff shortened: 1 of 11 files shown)


skeptrunedev (changes requested):
This needs frontend to be testable. W/out it's just too hard to figure out if things are working correctly.

frontends/search/src/components/ResultsPage.tsx
```diff
@@ -212,9 +212,16 @@ const ResultsPage = (props: ResultsPageProps) => {
...
+      const query =
+        props.search.debounced.multiQueries.length > 0
+          ? props.search.debounced.multiQueries
+              .map((q) => [q.query, q.weight])
+              .filter((q) => q[0] != "")
+          : props.search.debounced.query;
+
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       const requestBody: any = {
-        query: props.search.debounced.query,
```
drew-harris, line 217: you still would want the ResultsPage to use the debounced query so that it doesn't send double requests.

skeptrunedev: Above it's still using debounce, so going to resolve for now.

skeptrunedev (approved):

PR #2123, opened 2024-08-06
feature: clickhouse service to batch events

densumesh:


skeptrunedev (changes requested):
image

skeptrunedev (approved):
Nice latency improvement. Very fancy system

PR #2141, opened 2024-08-07
feature: support more distance metrics

cdxker:
Add support for dot product, manhattan distance, and euclidean distance metrics as a new dataset setting


densumesh (approved):
LGTM

PR #2144, opened 2024-08-08
feature: typo tolerance

aaryanpunia:
When spinning up services the amount of word_workers > bktree_workers. If dealing with big datasets spin up word_workers to 15+ to accommodate load CH can take the abuse

### DIFF
--- a/.github/workflows/push-server.yml
+++ b/.github/workflows/push-server.yml
@@ -450,3 +450,147 @@ jobs:
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+
+  word-id-cronjob:
+    name: Push Word ID cronjob
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      matrix:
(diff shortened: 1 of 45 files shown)


cdxker (changes requested):
Can we change id-worker to a better name and include `cronjob` instead of worker

aaryanpunia:
Pending problems - 

1. In word-worker, somehow, when inserting in the relation table, we are inserting the same (dataset_id, word_id) tuple multiple times in a single call to Postgres, this is not allowed when utilizing an on conflict do update.
2. Deadlocks with multiple word-workers, solvable by utilizing clickhouse instead of Postgres.

cdxker (changes requested):
Recursive data structure is too deep, getting "bk tree depth limit exceeded" serde errror when trying to load into the server. Will need to store the tree as a different structure. First thing that comes to mind is an adjacency list.

cdxker (approved):

cdxker (changes requested):
Can it return the corrected query on the response?

cdxker (approved):

PR #2162, opened 2024-08-12
feat: debuggin memory leak in highlights code, adding a flag to never inline

cdxker:

### DIFF
--- a/server/src/operators/chunk_operator.rs
+++ b/server/src/operators/chunk_operator.rs
@@ -1332,6 +1332,7 @@ pub fn get_slice_from_vec_string(vec: Vec<String>, index: usize) -> Result<Strin
     }
 }
 
+#[inline(never)]
 pub fn get_stop_words() -> Vec<String> {
     include_str!("../stop-words.txt")
         .lines()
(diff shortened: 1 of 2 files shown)


skeptrunedev (dismissed):

server/src/operators/chunk_operator.rs
```diff
@@ -1346,6 +1346,7 @@ pub enum HighlightStrategy {
     V1,
 }
 
+#[inline(never)]
 pub fn get_highlights_with_exact_match(
```
cdxker, line 1351: This is where I suspect the fishy code is at, but added traces at other spots just in case

server/src/operators/chunk_operator.rs
```diff
@@ -1332,6 +1332,7 @@ pub fn get_slice_from_vec_string(vec: Vec<String>, index: usize) -> Result<Strin
     }
 }
 
+#[inline(never)]
 pub fn get_stop_words() -> Vec<String> {
```
cdxker, line 1336: This also is a likely culprit, but we've had this function in the code base for a while so I doubt this will be the root cause.

PR #2164, opened 2024-08-12
bugfix: usize underfloat memory error, add timings to highlight_chunks

cdxker:
Added timing headers for highlight_chunks to measure perf on it (It was being bundled with the postgres span before).  and reverted the debug pr where we set debug = true.

The error seemed to be an underflow https://github.com/devflowinc/trieve/pull/2164/files#diff-bb28a6086cdf7de374c3cbf10a5271656a933be64adc38cfcfe414148b213ff1R1445-R1448 here

### DIFF
--- a/server/src/operators/chunk_operator.rs
+++ b/server/src/operators/chunk_operator.rs
@@ -1441,7 +1441,10 @@ pub fn get_highlights_with_exact_match(
             .to_string(),
     );
     let query_split = cleaned_query.split_whitespace().collect_vec();
-    let mut starting_length = query_split.len() - 1;
+    let mut starting_length = 0;
+    if !query_split.is_empty() {
+        starting_length = query_split.len() - 1;
+    }
     while starting_length > 0 {
         let mut current_skip = 0;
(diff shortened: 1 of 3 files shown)


server/src/operators/chunk_operator.rs
```diff
@@ -1443,7 +1441,10 @@ pub fn get_highlights_with_exact_match(
...
     );
     let query_split = cleaned_query.split_whitespace().collect_vec();
-    let mut starting_length = query_split.len() - 1;
+    let mut starting_length = 0;
+    if !query_split.is_empty() {
+        starting_length = query_split.len() - 1;
+    }
     while starting_length > 0 {
         let mut current_skip = 0;
         while current_skip <= query_split.len() - starting_length {
```
cdxker, line 1450: In production I think this wasn't erroring but instead underflowing to a very large `usize` value meaning that it allocated way more memory than we were expecting it to allocate. Causing it to be OOM KILLED

skeptrunedev (approved):

PR #2166, opened 2024-08-12
feature: add bm25 reranking, autocomplete, and recomendations

densumesh:


cdxker:
Went ahead and added in recommendations and added autocomplete to the search playground so people can find it easier.

cdxker (approved):

PR #2172, opened 2024-08-13
feature: get the count of items that have a tag associated with them

densumesh:

### DIFF
--- a/server/src/handlers/chunk_handler.rs
+++ b/server/src/handlers/chunk_handler.rs
@@ -2621,12 +2621,12 @@ pub fn check_completion_param_validity(
 }
 
 #[derive(Serialize, Deserialize, Debug, ToSchema)]
-pub struct GetItemsWithTagReqPayload {
+pub struct CountItemsWithTagReqPayload {
     pub tag: String,
 }
 
 #[derive(Serialize, Deserialize, Debug, ToSchema)]
-pub struct GetItemsWithTagResponse {
+pub struct CountItemsWithTagResponse {
     pub count: i64,
 }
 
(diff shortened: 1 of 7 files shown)


server/src/handlers/chunk_handler.rs
```diff
@@ -2619,3 +2619,43 @@ pub fn check_completion_param_validity(
...
+
+#[derive(Serialize, Deserialize, Debug, ToSchema)]
+pub struct GetItemsWithTagReqPayload {
+    pub tag: String,
+}
+
+#[derive(Serialize, Deserialize, Debug, ToSchema)]
+pub struct GetItemsWithTagResponse {
+    pub count: i64,
+}
```
skeptrunedev, line 2631: Feels like these are names poorly given all this route does is return a count

server/src/handlers/chunk_handler.rs
```diff
@@ -2619,3 +2619,43 @@ pub fn check_completion_param_validity(
...
+    _user: LoggedUser,
+    pool: web::Data<Pool>,
+    dataset_org_plan_sub: DatasetAndOrgWithSubAndPlan,
+) -> Result<HttpResponse, ServiceError> {
+    let dataset_id = dataset_org_plan_sub.dataset.id;
+
+    let count = get_items_with_tag_query(tag.into_inner().tag, dataset_id, pool).await?;
+
+    Ok(HttpResponse::Ok().json(GetItemsWithTagResponse { count }))
+}
```
skeptrunedev, line 2661: Does it make sense to change the `path` here and respond with the chunks correlating to the tag **and** the count?

skeptrunedev (approved):

skeptrunedev:
Closing for now since the `chunk/count` route can be used for this purpose

PR #2176, opened 2024-08-13
feature: get all tags for a dataset with their counts

densumesh:


server/src/handlers/dataset_handler.rs
```diff
@@ -551,3 +551,55 @@ pub async fn get_datasets_from_organization(
...
+}
+
+#[utoipa::path(
+    post,
+    path = "/dataset/get_all_tags",
+    context_path = "/api",
+    tag = "Dataset",
+    request_body(content = GetAllTagsReqPayload, description = "JSON request payload to get items with the tag in the request", content_type = "application/json"),
+    responses(
+        (status = 200, description = "Count of the items with the tag that you were searching for", body = GetAllTagsResponse),
```
skeptrunedev, line 580:
```suggestion
        (status = 200, description = "Page of tags requested with all tags and the number of chunks in the dataset with that tag plus the total number of unique tags for the whole datset", body = GetAllTagsResponse),
```

server/src/handlers/dataset_handler.rs
```diff
@@ -551,3 +551,55 @@ pub async fn get_datasets_from_organization(
...
+    )
+)]
+pub async fn get_all_tags(
+    data: web::Json<GetAllTagsReqPayload>,
+    _user: LoggedUser,
+    pool: web::Data<Pool>,
+    dataset_org_plan_sub: DatasetAndOrgWithSubAndPlan,
+) -> Result<HttpResponse, ServiceError> {
+    let dataset_id = dataset_org_plan_sub.dataset.id;
+    let page = data.page.unwrap_or(1);
```
skeptrunedev, line 597: Something here should handle it potentially being 0

server/src/operators/dataset_operator.rs
```diff
@@ -629,3 +630,41 @@ pub async fn get_dataset_usage_query(
...
+    let mut conn = pool.get().await.unwrap();
+
+    let items = dataset_tags_columns::dataset_tags
+        .inner_join(chunk_metadata_tags_columns::chunk_metadata_tags)
+        .group_by(dataset_tags_columns::tag)
+        .select((
+            dataset_tags_columns::tag,
+            count(chunk_metadata_tags_columns::chunk_metadata_id),
+        ))
+        .order_by(count(chunk_metadata_tags_columns::chunk_metadata_id).desc())
```
skeptrunedev, line 652: I'm worried about perf, but we'll test on HN. 

I'm pretty sure this forces it to calc the count for every tag in the whole dataset.

server/src/operators/dataset_operator.rs
```diff
@@ -629,3 +630,41 @@ pub async fn get_dataset_usage_query(
...
+        .map_err(|err| {
+            ServiceError::BadRequest(format!("Failed to get items with tags {}", err))
+        })?;
+
+    let total_count = dataset_tags_columns::dataset_tags
+        .select(count(dataset_tags_columns::tag))
+        .filter(dataset_tags_columns::dataset_id.eq(dataset_id))
+        .first::<i64>(&mut conn)
+        .await
+        .map_err(|err| ServiceError::BadRequest(format!("Failed to get count of tags {}", err)))?;
```
skeptrunedev, line 667: These should probably in a futures::join

skeptrunedev: But again, will test latency on HN and see if refactor is needed. Not going to do it rn when the entire feature is up in the air latency-wise.

skeptrunedev (approved):

PR #2178, opened 2024-08-13
feature: add dataset button to analytics

vid277:
image


drew-harris:
I would move the dashboard icon to the bottom of that box to show that it's on a separate site. It looks like just another page on the analytics when its not.  Could also but a link to the api docs there too.

drew-harris:
Like here
image

vid277:
image

.env.example
```diff
@@ -24,6 +24,11 @@ VITE_DASHBOARD_URL="http://localhost:5173"
 ##### SEARCH #####
 VITE_SENTRY_SEARCH_DSN=""
 
+##### ANALYTICS #####
+VITE_API_HOST=http://localhost:8090/api
+VITE_SEARCH_UI_URL="http://localhost:5174"
+VITE_DASHBOARD_UI_URL="http://localhost:5173"
+
```
cdxker, line 31:
```suggestion
```
These shouldn't be needed, as they are already defined in other locations.

cdxker (changes requested):
One comment request and can you change the env name to `VITE_DASHBOARD_URL`, just to be consistent with how its named in the search ui and dashboard UI. Other than that it looks great!

densumesh (approved):

PR #2198, opened 2024-08-15
fix: fix fetch client build script

drew-harris:


drew-harris:
Added fixup commit to fix type error

PR #2211, opened 2024-08-16
feature: use hashmap to limit rtt on reqeusts to get group tracking ids

cdxker:
Use a hashmap at the begining of the request and make any groups and great the groups at the top of the batch.

### DIFF
--- a/server/src/operators/chunk_operator.rs
+++ b/server/src/operators/chunk_operator.rs
@@ -2184,6 +2184,53 @@ pub async fn create_chunk_metadata(
 
     let mut chunk_metadatas = vec![];
 
+    let mut group_tracking_ids_to_group_ids: HashMap<String, uuid::Uuid> = HashMap::new();
+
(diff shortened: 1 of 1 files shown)


server/src/operators/chunk_operator.rs
```diff
@@ -2184,6 +2184,53 @@ pub async fn create_chunk_metadata(
...
+                    None,
+                    None,
+                ))
+            } else {
+                None
+            }
+        })
+        .collect();
+
+    let created_groups = create_groups_query(new_groups, true, pool.clone()).await?;
```
cdxker, line 2226: Also setting upsert_by_group_tracking_id to `true` just in case of the race condition that 2 servers are trying to create the group at the same time. We want them to agree on the same id, onConflictDoNothing will not return a group_id at all which might cause some issues.

skeptrunedev (approved):

PR #2237, opened 2024-08-23
feature: add `tracking_id` and `id` filter on search dashboarad

vid277:


skeptrunedev (approved):
Both new filter types work great! Super awesome feature

PR #2245, opened 2024-08-24
bugfix: chunks get inserted in completion_first manner when completion_first = true

cdxker:


server/src/handlers/message_handler.rs
```diff
@@ -247,8 +247,23 @@ pub async fn get_all_topic_messages(
...
-    let messages =
-        get_messages_for_topic_query(topic_id, dataset_org_plan_sub.dataset.id, &pool).await?;
+    let messages: Vec<models::Message> = get_messages_for_topic_query(topic_id, dataset_org_plan_sub.dataset.id, &pool)
+        .await?
+        .into_iter()
+        .filter_map(|mut message| {
+            if message.content[0..4] == *"||[{" {
+                match message.content.rsplit_once("}]") {
+                    Some((chunks, ai_message)) => {
+                        message.content = format!("{}}}]{}", chunks, ai_message);
```
cdxker, line 257: In the format string `}}` = a single `}`. Not sure why they don't just do a `\}` but yeah.

server/src/operators/message_operator.rs
```diff
@@ -611,12 +611,25 @@ pub async fn stream_response(
...
+        .map(|x| x.completion_first)
+        .unwrap_or(Some(false))
+        .unwrap_or(false);
+
     Arbiter::new().spawn(async move {
         let chunk_v: Vec<String> = r.iter().collect();
         let completion = chunk_v.join("");
 
+        let message_to_be_stored = if completion_first {
+            format!("{}{}", completion, chunk_metadatas_stringified)
```
cdxker, line 626: I think it makes the most sense to store the chunks in the completion order the user wanted them to be streamed in. (Will likely work a lot better for the front ends to be modular.

server/src/handlers/message_handler.rs
```diff
@@ -247,8 +258,23 @@ pub async fn get_all_topic_messages(
...
 
-    let messages =
-        get_messages_for_topic_query(topic_id, dataset_org_plan_sub.dataset.id, &pool).await?;
+    let messages: Vec<models::Message> =
+        get_messages_for_topic_query(topic_id, dataset_org_plan_sub.dataset.id, &pool)
+            .await?
+            .into_iter()
+            .filter_map(|mut message| {
+                if message.content.starts_with("||[{") {
+                    match message.content.rsplit_once("}]") {
```
cdxker, line 267: In the format string }} = a single }. Not sure why they don't just do a \} but yeah.

server/src/handlers/message_handler.rs
```diff
@@ -471,12 +497,23 @@ pub async fn regenerate_message_patch(
...
+                        .to_string();
+                } else {
+                    // This is (content, chunks)
+                    message.content = message
+                        .content
+                        .rsplit("||")
+                        .last()
+                        .unwrap_or("I give up, I can't find a citation")
+                        .to_string();
+                }
```
cdxker, line 516: This should with a strange combo of completion_first = true / false

skeptrunedev (commented):
Looks great! Tested by cherry-pick'ing the frontend commit and duplicating the error then validating that this fixes it.

skeptrunedev (approved):

PR #2254, opened 2024-08-26
feat: organization name prompt for new users

drew-harris:

### DIFF
--- a/server/src/handlers/auth_handler.rs
+++ b/server/src/handlers/auth_handler.rs
@@ -1,5 +1,4 @@
 use crate::data::models::{Organization, RedisPool, StripePlan, UserRole};
-use crate::data::schema::user_organizations;
 use crate::get_env;
 use crate::operators::invitation_operator::check_inv_valid;
 use crate::operators::organization_operator::{get_org_from_id_query, get_user_org_count};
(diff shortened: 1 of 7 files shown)


drew-harris:
image

SaraVieira (approved):
Can't vouch for the rust but FE looks good

skeptrunedev (changes requested):
I really don't agree with this being a page that blocks access to the entire dashboard. It should be implemented as a fullscreen modal which appears after you get to dash. If the user dismisses it then we should use the default org name. 

image

server/src/handlers/auth_handler.rs
```diff
@@ -287,6 +287,8 @@ pub struct AuthQuery {
     pub organization_id: Option<uuid::Uuid>,
     /// URL to redirect to after successful login
     pub redirect_uri: Option<String>,
+    /// URL to redirect to if the user is new
+    pub new_user_redirect_uri: Option<String>,
```
densumesh, line 291: I think instead of adding it like this, if it is a new user we should modify the rust code to add a query param to the redirect that indicates a new user was created

densumesh (dismissed):

skeptrunedev (approved):

skeptrunedev:
This is a really nice feature

PR #2261, opened 2024-08-27
feat - new search react component

SaraVieira:
- [x] Search component
- [x] Modal Search Component
- [ ] Add web components support

Example: https://trieve-react-search.surge.sh/


skeptrunedev (changes requested):
image

Major nit I have is that it should show the title then, subsequently, the highlights below it. If the title is in the chunk_html then it should be removed as well. I would look for <b><mark> children matching the title text and remove those then mark the title as highlighted. But also maybe don't remove? I'm not sure, should study how Mintlify does it.

skeptrunedev:
I think we should highly prioritize setting up chat for this as well

SaraVieira:
The highlights should be much better now and will work on chat tomorrow

skeptrunedev (approved):
This is awesome. Definitely a fully functional component for search and I'm hype about it. Arrows and everything work and the display looks great. 

Sorry for being pedantic with the commit history, but I'm nervous our total commit count is getting too large and makes us seem sus lol. Hype about this!

Issue #2181, opened 2024-08-14
tracking: glasskube package


cdxker:
### Description

Wanting to migrate from helm exclusivley to Glasskube. Mainly for easier management of different custom releases and   having databases be configured better via the helm chart

- [x] Refactor helm chart to remove database. (Swap to using external Glasskube packages) #2266
    - [x] Qdrant Glasskube packges
- [x] security: configure secret references to pull from databases (instead of envs / values.yaml) #2266 
    - [x] security: onfigure keycloak realm and push OIDC SECRET into secret that `server` references
    - [x] security: setup other server secretmaps (postgres, clickhouse, s3, etc)
    - [x] cleanup: server and workers share the same configmap. (easier to maintain instead of copy/pasting envs)
- [ ] Deployment process
     - [x] Trieve release bot
     - [x] Add release please bot
     - [x] Tagged docker images based on git tag
     - [ ] <del> helm chart "safe enough" to be pushed to Dockerhub </del>
        - [ ] <del> helm chart build/push CI action </del>
- [ ] <del> Create Glasskube package pull request. Add to docs. (For minikube cpu based setup). </del>
- [ ] <del> docs: update documentation for how to release via glasskube </del>

### Community channels

Matrix is preferred. Reach out on discord or Matrix for further assistance. 

- [discord](https://discord.gg/CuJVfgZf54)
- [matrix](https://matrix.to/#/#trieve-general:trieve.ai)

pmig:
Open tasks:

- [ ] refactor workers to base worker as they share most of the the configuration.

PR #2266, opened 2024-08-30
feat: add glasskube installation instructions

pmig:
ref: #2181

Rewrite Kubernetes deployment via kustomize and add docs.


cdxker (approved):
A great starting out template, merging so that we can have smaller PR's in the future!

PR #2269, opened 2024-08-30
fix: remove tooltip in chat to fix cutoff

drew-harris:


drew-harris:
Couldn't get the tooltip to stop overflowing cuz of the resizable so i just used html title tag for popup instead. Should be fine since all the other buttons on the frontend use that instead.

skeptrunedev (approved):
This was some **ancient** tech debt. In Arguflow times we had the concept of a public and private chunk since it was designed more for an enterprise RAG use-case. That has since gone away as we are more of an API such that this entire feature can be dumped. Fully removed it and fixed a small lint in server code.

PR #2271, opened 2024-08-30
fix: mintlify boldification

drew-harris:

Temporary solution until I can write a vite plugin that wipes out the tailwind reset in the widget.


skeptrunedev (approved):
Easy win. Should keep in mind that this file probably needs to get updates as Mintlify updates their component.

PR #2273, opened 2024-08-31
feature: recognize domain specific words

densumesh:


skeptrunedev (commented):
Are the dependency removals here intentional?

densumesh:
> Are the dependency removals here intentional? 

Yes they are. No longer using bloom filters and bktree functions have been moved into our code

server/src/operators/typo_operator.rs
```diff
@@ -650,13 +660,20 @@ fn correct_query_helper(tree: &BkTree, query: String, options: &TypoOptions) ->
...
 
     if corrections.is_empty() {
-        None
+        CorrectedQuery {
+            query: Some(query),
+            corrected: false,
+        }
     } else {
-        let mut corrected_query = query.to_string();
+        let mut corrected_query = query.query.clone();
```
skeptrunedev, line 668: Behavior is correct, but it shouldn't always happen. `TypoOptions` needs another param, which defaults to True, such that the user can turn this behavior off if they want to.

skeptrunedev (changes requested):

skeptrunedev (changes requested):
new request param needs to get added to options popover

skeptrunedev (approved):

PR #2276, opened 2024-09-01
feat: add clickhouse embed function on initdb

pmig:

@cdxker Given the example of user defined function (https://github.com/Altinity/clickhouse-operator/blob/master/docs/chi-examples/23-udf-example.yaml) shouldn't we able to make all the config changes of https://github.com/devflowinc/trieve/blob/main/docker/clickhouse/Dockerfile also in the custom resource?

Then we can use the original clickhouse image and make updates easier.

### DIFF
--- a/glasskube/trieve/clickhouse/clickhouse.yaml
+++ b/glasskube/trieve/clickhouse/clickhouse.yaml
@@ -20,6 +20,11 @@ spec:
     podTemplates:
       - name: clickhouse-pod-template
         spec:
+          volumes:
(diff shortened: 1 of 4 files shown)


cdxker:
Yes @pmig I think it would be far more ideal. Can you set that up? After I can hop on a call with @densumesh to ensure all the custom functionality he wants is in that docker image.

pmig:
@cdxker sounds good, will update the PR!

pmig:
superceeded by https://github.com/devflowinc/trieve/pull/2290

Issue #2226, opened 2024-08-21
feat: indicate in chunk container top-right corner buttons when in chunk view


danielsgriffin:
### Description

Chunk containers (in search) display four buttons on the top-right corner: chunk view, chunk/edit, delete, manage groups
This is the tooltip for the chunk view button: 
<img width="452" alt="image" src="https://github.com/user-attachments/assets/5ec3a056-8706-49c4-94a3-4cf44c849903">
The button and tooltip remain the same whether the user is already in chunk view or not, leading to confusion if the user doesn't recognize that is already where they are. This might happen particularly if they click to view a chunk from the chat docs.

### Target(s)

search

### Requirement to close

PR which provides some visual indicator on the button that the current view is the chunk view.

### Community channels

Matrix is preferred. Reach out on discord or Matrix for further assistance. 

- [discord](https://discord.gg/CuJVfgZf54)
- [matrix](https://matrix.to/#/#trieve-general:trieve.ai)

PR #2281, opened 2024-09-03
fix: hide open recs icon when already viewing recs

drew-harris:


drew-harris:
fixes #2226

skeptrunedev (approved):
nice cleanup on this component in the single page view

PR #2283, opened 2024-09-03
feature: fix casing issue and pull from redis in the foreground

densumesh:


skeptrunedev (approved):
NIIIICEEE

`FreeBSD` is going to work

PR #2295, opened 2024-09-06
[Search Component] - CSS Scoping

SaraVieira:
- Adds better css scoping so we don't get overriden by the users CSS
- Bumps version


skeptrunedev (approved):

skeptrunedev (changes requested):
```
cleanup: better CSS scoping for search component
chore: regenerate lock
```

skeptrunedev (approved):

PR #2314, opened 2024-09-09
fix: css fixes for rag page

SaraVieira:
- Removes select that returns 500 when trying to use
- Adds table headers
- Makes top padding more consistent

Before
image

After:
image

Next PR will add a modal showing results


skeptrunedev (changes requested):
Somewhat surprised to see `mintlify.js` file here. Are we sure it should be getting included?

SaraVieira:
That was definitly a mistake, fixed

skeptrunedev (approved):

PR #2321, opened 2024-09-10
feat: new tables for rag queries

SaraVieira:
- Adds tanstack table to rag queries
- Formats table with new style
- Adds more info to table

Next PR:
- Redesign the way to show the results

image

### DIFF
--- a/frontends/analytics/src/components/Pagination.tsx
+++ b/frontends/analytics/src/components/Pagination.tsx
@@ -1,0 +1,51 @@
+import { usePagination } from "../hooks/usePagination";
+import { IoChevronBack, IoChevronForward } from "solid-icons/io";
+
+interface PaginationProps {
+  pages: ReturnType<typeof usePagination>;
+  total?: number;
+  perPage?: number;
+}
+export const Pagination = (props: PaginationProps) => {
+  return (
+    <nav
+      class="flex items-center justify-between border-t border-gray-200 bg-white py-3 sm:px-6"
+      aria-label="Pagination"
+    >
+      {props.perPage && props.total ? (
+        <div class="hidden sm:block">
+          <p class="text-sm text-gray-700">
+            Showing
+            <span class="px-1 font-medium">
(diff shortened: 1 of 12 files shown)


frontends/analytics/src/components/Pagination.tsx
drew-harris, line 1: This would be nice to have in the shared component library

frontends/analytics/src/components/Table.tsx
drew-harris, line 1: This one should also be in the shared library i think

frontends/analytics/src/components/charts/RagQueries.tsx
```diff
@@ -84,40 +152,53 @@ export const RagQueries = (props: RagQueriesProps) => {
...
+              },
+            },
+            columns: columns(),
+            getCoreRowModel: getCoreRowModel(),
+            manualPagination: true,
+          });
+          const usage = queryClient.getQueryData<RAGUsageResponse>([
+            "rag-usage",
+            { filter: props },
+          ]);
```
drew-harris, line 177: Feels like this should just be in the main body of the component function. We have so much logic in the rendering on the other sites and its super messy. Would like to avoid that on analytics lol

SaraVieira: Thats super fair, I tried that first and stuff just wouldn't update, let me give it another try

frontends/analytics/src/components/Table.tsx
SaraVieira: Done! Named it TanStack Table because we already have the other table, let me know if that makes sense

frontends/analytics/src/components/Pagination.tsx
SaraVieira: Done!

skeptrunedev:
Commits look like something went wrong as there are some from the main branch present. I think the easiest thing to do would be collapse down into one commit which is what I'm leaning towards on merge.

skeptrunedev (approved):

PR #2323, opened 2024-09-11
feature: set up scrape worker with endpoint to crawl

densumesh:
New ENVs:
FIRECRAWL_URL: defaults to cloud impl
FIRECRAWL_API_KEY
Export-1726702687599
Export-1726606212948


skeptrunedev (changes requested):
I think this is architected wrong. We need to default to a behavior where we auto-crawl on a cron instead of requiring the trigger.

server/src/handlers/chunk_handler.rs
```diff
@@ -2674,3 +2675,73 @@ pub fn check_completion_param_validity(
...
+        data.site,
+        dataset_org_plan_sub.dataset.id,
+        scrape_id,
+        interval,
+        pool,
+        redis_pool,
+    )
+    .await?;
+    Ok(HttpResponse::Ok().json(CrawlResponse { scrape_id }))
+}
```
skeptrunedev, line 2747: We should move this to something that happens on dataset creation and update instead of having it be its own route.

I tend to think this is something people will almost exclusively do via the dashboard UI.

skeptrunedev (changes requested):

server/src/handlers/dataset_handler.rs
```diff
@@ -185,6 +209,10 @@ pub struct UpdateDatasetRequest {
     pub server_configuration: Option<DatasetConfigurationDTO>,
     /// Optional new tracking ID for the dataset. Can be used to track the dataset in external systems. Must be unique within the organization. If not provided, the tracking ID will not be updated. Strongly recommended to not use a valid uuid value as that will not work with the TR-Dataset header.
     pub new_tracking_id: Option<String>,
+    /// New site to crawl for the dataset. If provided, the dataset will be populated with the contents of the site.
+    pub crawl_site: Option<String>,
+    /// New crawl interval for the site. If provided, the dataset will be populated with the contents of the site on a recurring basis.
```
skeptrunedev, line 214: Should have the default interval

skeptrunedev (changes requested):

server/src/handlers/dataset_handler.rs
```diff
@@ -107,6 +115,8 @@ pub struct CreateDatasetRequest {
 pub async fn create_dataset(
     data: web::Json<CreateDatasetRequest>,
     pool: web::Data<Pool>,
+    redis_pool: web::Data<RedisPool>,
+    dataset_org_plan_sub: DatasetAndOrgWithSubAndPlan,
```
skeptrunedev, line 119: This doesn't work since we can't have a `TR-Dataset` header before the dataset is created

skeptrunedev (changes requested):

skeptrunedev (changes requested):
- [ ] indicate class/id selector for which div to pull content out of
- [ ] `<heading>` or `h` tag will be boost phrase` to boost for both fulltext and semantic by 1.3 0.3

skeptrunedev (changes requested):
image

Doesn't seem like the `include_tags` and `exclude_tags` are being passed to firecrawl properly.

server/src/operators/crawl_operator.rs
```diff
@@ -375,37 +378,85 @@ pub async fn update_scrape_id(
...
+        };
+
+        let ingest_result = response.json::<IngestResult>().await.map_err(|e| {
             log::error!("Error parsing response from firecrawl: {:?}", e);
             ServiceError::InternalServerError("Error parsing response from firecrawl".to_string())
         })?;
+        if ingest_result.status != Status::Completed {
+            log::info!("Crawl status: {:?}", ingest_result.status);
+            return Ok(ingest_result);
+        }
```
densumesh, line 421: need to capture the failed status as well or else this will error out

skeptrunedev (approved):

PR #2337, opened 2024-09-13
fix: assorted fixes on component from signoz feedback

SaraVieira:
- [x] Add a11y info at the bottom
- [x] Add border to the footer
- [x] Add possibility to pass custom button component to open modal
- [x] use autcomplete instead of search
- [x] sendCTR analaytics on result click
- [x] z-index fixes


skeptrunedev (approved):
Need to publish a new package following merge.

PR #2348, opened 2024-09-18
fix: add delete invitation to openapi spec

drew-harris:


drew-harris:
Looks like including it creates an error about ambiguous paths. Will look into this

skeptrunedev:
> Looks like including it creates an error about ambiguous paths. Will look into this

Got it sorted. Can't have `DELETE /api/{foo}` and `GET /api/{bar}`

skeptrunedev (approved):

PR #2369, opened 2024-09-21
feature: add refresh chunk count button 

vid277:
Add refresh chunk count button to dataset start and dataset settings page.
Also fixed the copy dataset id button.

### DIFF
--- a/frontends/dashboard/src/components/DatasetOverview.tsx
+++ b/frontends/dashboard/src/components/DatasetOverview.tsx
@@ -22,17 +22,21 @@ import {
   createSolidTable,
   getCoreRowModel,
 } from "@tanstack/solid-table";
-import { DatasetAndUsage } from "trieve-ts-sdk";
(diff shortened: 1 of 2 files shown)


frontends/dashboard/src/components/DatasetOverview.tsx
```diff
@@ -48,6 +50,57 @@ export const DatasetOverview = () => {
...
+          "Content-Type": "application/json",
+        },
+        credentials: "include",
+      });
+
+      if (!response.ok) {
+        throw new Error("Failed to fetch dataset usage");
+      }
+
+      const newData = await response.json();
```
skeptrunedev, line 72:
```suggestion
      const newData = await response.json() as {insert correct type from SDK types here}
```

Lint errors are coming because this type is unkown.

skeptrunedev (changes requested):

skeptrunedev:
image

skeptrunedev (approved):

PR #2371, opened 2024-09-21
cleanup: remove eslint_report.json from repo

skeptrunedev:


cdxker (approved):
yeah the ci makes this file, we should remove it from the repo as it is not needed for development

Issue #2357, opened 2024-09-19
feature: search component query suggestions


skeptrunedev:
### Description

- [ ] add fields for request payload options of query suggestions with sensible defaults to component
- [ ] add refresh button to query suggestions if the user is using the route
- [ ] add ability to manually spec query suggestion if user does not want to use the route

### Target(s)

`search-component`

### Requirement to close

PR checking the 3 boxes in description

### Community channels

Matrix is preferred. Reach out on discord or Matrix for further assistance. 

- [discord](https://discord.gg/CuJVfgZf54)
- [matrix](https://matrix.to/#/#trieve-general:trieve.ai)

PR #2383, opened 2024-09-23
feat: add suggested queries to search component

SaraVieira:
- [x] Add suggested queries to the UI
- [x] Move all prop drilling to context
- [x] Allow user to say they don't want suggested queries
- [x] Allow user to pass default search queries
- [x] show refresh button only on empty state
- [x] only show on empty state or no results
image

image

closes https://github.com/devflowinc/trieve/issues/2357
closes https://github.com/devflowinc/trieve/issues/2392


skeptrunedev:
- refresh button
- only show suggestions on empty state

skeptrunedev (changes requested):
image
image

- [ ] should show 3 suggestions instead of 2 by default
- [ ] loading animation once the first results load in should just `animate-pulse` instead of having the larger dots
- [ ] loading animation on first try should say something like `Loading random query suggestions...` with an `animate-pulse` so user knows what's happening

skeptrunedev (approved):

PR #2385, opened 2024-09-23
feat: add editMessageReader and regenerateMessageReader methods

Yonom:
Adding editMessageReader and regenerateMessageReader methods analogous to createMessageReader
This allows for streaming results after editing or regenerating a message


skeptrunedev (changes requested):
Going to pull this PR real quick and generate a new client off it so we can republish.

skeptrunedev (approved):

PR #2395, opened 2024-09-24
feat: analytics on dashboard

drew-harris:


densumesh:
image

densumesh (changes requested):
need to add back button

frontends/dashboard/src/analytics/components/charts/NoResultQueries.tsx
```diff
@@ -73,18 +74,15 @@ export const NoResultQueries = (props: NoResultQueriesProps) => {
   console.log(notResultQuery.data);
```
SaraVieira, line 74: console log left behind

SaraVieira:
Besides the back button mentioned the only think I have to add is that the button to open the analytics is still in the menu

Other than that this looks great!

densumesh (approved):

PR #2405, opened 2024-09-25
feat(glasskube): add aws package overlay

pmig:


cdxker (approved):
This looks good, merging so we can make more PR's off of this later.

PR #2419, opened 2024-09-26
feat: better org switching

drew-harris:


drew-harris:
Going to close until I can get a better popover component made or make a pr to terracotta. All functionality this pr would add already exists in other components

PR #2420, opened 2024-09-26
feature: get custom events with a metadata filter

densumesh:


cdxker (changes requested):
Needs a version bump as this is a breaking change for people using the the REST endpoints directly.

Issue #2427, opened 2024-09-27
feature: get CSV exports of analytics data


skeptrunedev:
### Description

export analytics data such that queries like "how many time people are going for search result and how many times people are going for AI generated result?" can be answered. 

### Target(s)

`analytics`

### Requirement to close

PR enabling the functionality described in the description above is merged.

### Community channels

Matrix is preferred. Reach out on discord or Matrix for further assistance. 

- [discord](https://discord.gg/CuJVfgZf54)
- [matrix](https://matrix.to/#/#trieve-general:trieve.ai)

PR #2432, opened 2024-09-27
Feat: CSV Export of analytics table

SaraVieira:
- [x] create general function in the table that go through all pages and gathers the data
- [x] create and download CSV
- [x] Add this to all tables

image

closes https://github.com/devflowinc/trieve/issues/2427


SaraVieira:
Questions: 

- Should I just do an export of all the data that comes back from the server or just create a CSV of the shown columns? Right now it's doiing the first one

densumesh:
> Questions:
> 
> * Should I just do an export of all the data that comes back from the server or just create a CSV of the shown columns? Right now it's doiing the first one

i think it should be getting all of the data from the server

densumesh (approved):

PR #2435, opened 2024-09-27
feat: set tracking ids for datasets

drew-harris:

### DIFF
--- a/server/src/operators/dataset_operator.rs
+++ b/server/src/operators/dataset_operator.rs
@@ -570,7 +570,13 @@ pub async fn update_dataset_query(
             .filter(datasets_columns::deleted.eq(0)),
     )
     .set((
-        new_tracking_id.map(|id| datasets_columns::tracking_id.eq(id)),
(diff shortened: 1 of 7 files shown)


cdxker:
Added in a backend PR that sets an  empty tracking_id string to just delete instead, no we don't have to worry about empty states.

server/src/operators/dataset_operator.rs
```diff
@@ -570,7 +570,13 @@ pub async fn update_dataset_query(
...
     )
     .set((
-        new_tracking_id.map(|id| datasets_columns::tracking_id.eq(id)),
+        new_tracking_id.map(|id| {
+            if id == "" {
+                datasets_columns::tracking_id.eq(None)
+            } else {
+                datasets_columns::tracking_id.eq(Some(id))
+            }
+        }),
```
cdxker, line 579: Added in a backend part to this that sets an empty tracking_id string to just delete instead, no we don't have to worry about empty states.

cdxker (approved):

PR #2449, opened 2024-09-29
feature: mobile view for search component

cdxker:
mobile_ui

UI still needs fix for auto scroll to work properly. Will need to be fixed in another PR.

### DIFF
--- a/clients/search-component/src/utils/hooks/useChat.ts
+++ b/clients/search-component/src/utils/hooks/useChat.ts
@@ -44,6 +44,14 @@ export const useChat = () => {
       if (doneReading) {
         done = doneReading;
         setIsLoading(false);
+
+        // Get First Chat Container 
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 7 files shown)


clients/search-component/src/utils/hooks/useChat.ts
```diff
@@ -44,6 +44,14 @@ export const useChat = () => {
       if (doneReading) {
         done = doneReading;
         setIsLoading(false);
+
+        // Get First Chat Container 
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const chatContainer: any = modalRef.current?.querySelector("div.chat-container");
+        if (chatContainer) {
+          chatContainer.style.height = "fit-content";
+        }
```
cdxker, line 53: Had to do a hack here to get the footer to be at the right position after a message is generated.

skeptrunedev (changes requested):
Spacing between user and assistant blocks is off. They also need a border drawn between them.

cdxker:
image

Evened out the spacing.

cdxker:
image

Added vertical bars

PR #2452, opened 2024-09-29
feat: add dataset event types for crawl success/fail

drew-harris:


server/src/bin/crawl-worker.rs
```diff
@@ -84,12 +95,16 @@ async fn crawl(
 
     log::info!("Processing {} documents from scrape", data.len());
 
+    let mut page_count = 0;
```
densumesh, line 98: Make this data.len() and store the chunks created as well

densumesh (changes requested):

densumesh (approved):

PR #2465, opened 2024-10-01
feature: add rate rag route

densumesh:


server/src/operators/analytics_operator.rs
```diff
@@ -1382,6 +1382,49 @@ pub async fn set_query_rating_query(
     Ok(())
 }
 
+pub async fn set_rag_query_rating_query(
+    data: RateQueryRequest,
+    dataset_id: uuid::Uuid,
+    clickhouse_client: &clickhouse::Client,
```
cdxker, line 1388: Can you add this to openapi spec.

You might also need to regenerate the client

cdxker (changes requested):

server/src/operators/analytics_operator.rs
cdxker: Looking at the wrong spot

cdxker (approved):

PR #2467, opened 2024-10-01
feature: clear chat history in search component

cdxker:
clear-chat

### DIFF
--- a/.github/workflows/update-search-component.yaml
+++ b/.github/workflows/update-search-component.yaml
@@ -27,10 +27,12 @@ jobs:
           cache: yarn
 
       - name: Install dependencies 📦️
+        working-directory: ./clients/search-component/
         run: yarn --frozen-lockfile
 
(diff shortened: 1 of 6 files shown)


skeptrunedev (changes requested):
Flip the sides of powered and clear buttons

SaraVieira:
@skeptrunedev If we switch the sides it's gonna look off on search. I would just make the button less prominent, add a border instead of a bg color

skeptrunedev:
> @skeptrunedev If we switch the sides it's gonna look off on search. I would just make the button less prominent, add a border instead of a bg color

Sounds like a great plan to me

SaraVieira:
image

Made some small chanmges

- Button only appears when there are messages
- Added colors for dark mode
- Removed bg color and added a border instead

skeptrunedev (approved):
LGTM

Issue #2468, opened 2024-10-01
feature: clear chat button turns to cancel when generating chunks.


cdxker:
### Description

The clear chat button turns to cancel when generating chunks, on press chunk generating is canceled.

### Target(s)

`docsearch`

### Community channels

Matrix is preferred. Reach out on discord or Matrix for further assistance. 

- [discord](https://discord.gg/CuJVfgZf54)
- [matrix](https://matrix.to/#/#trieve-general:trieve.ai)

PR #2471, opened 2024-10-01
feature: Add stop generating button to chat component

SaraVieira:
Shows stop generating component in place of the clar chat while the AI is generating content

closes https://github.com/devflowinc/trieve/issues/2468


cdxker (changes requested):
In the loading state it seems to bug out. Can you fix this.
 
fix-loading-state

PR #2473, opened 2024-10-01
cleanup: hide footer when on mobile device

vid277:
image


clients/search-component/src/TrieveModal/index.tsx
```diff
@@ -20,6 +20,15 @@ const Modal = () => {
...
   const ButtonEl = props.ButtonEl;
 
+  const [isMobile, setIsMobile] = useState(false);
+
+  useEffect(() => {
+    const checkWindowSize = () => setIsMobile(window.innerWidth <= 768);
+    window.addEventListener("resize", checkWindowSize);
+    checkWindowSize();
+    return () => window.removeEventListener("resize", checkWindowSize);
+  }, []);
```
SaraVieira, line 30: We can use tailwind classes for this, https://tailwindcss.com/docs/responsive-design

PR #2512, opened 2024-10-02
bugfix: use rag queries for results

cdxker:
If #2511 does not fix the error, we  can default swap to rag_queries.results if there is no easy fix.


cdxker:
#2511 Did not fix the root issue, I suspect that the root issue has to do with serde serialization errors. Pushing this fix to at least show more data for the user.

densumesh (approved):

PR #2514, opened 2024-10-02
feature: send to dittofeed on new user

densumesh:


cdxker (changes requested):
Can you also do after dataset is created

cdxker (approved):

PR #2517, opened 2024-10-03
feat: hackernews trieve glasskube migration

pmig:

PR #2518, opened 2024-10-03
feat: trieve gcp package for hackernews deployment

pmig:
superced  https://github.com/devflowinc/trieve/pull/2517

Open points:

- increase slade-query to 1GB (instead of 500mb)


cdxker (changes requested):
Looks like the keycloak Statefulset isn't able to startup here, even thought the keycloak pod is available. (Must be a healthcheck issue) or some issue with a secret that isn't properly configured for the keycloak-operator.

cdxker (approved):
Will need to update the `ConfigMap` eventually, but this works for gcp

PR #2522, opened 2024-10-03
feature: fix serde deserialize issue

densumesh:


cdxker (changes requested):
This does not fix the error. thinking we have to fix it from the read side. I have this regex query that is a decent start to this. Going to finish this before I go to sleep.

```
"chunk_html":"(.*?)"metadata":
```

cdxker:
solved by #2534

PR #2526, opened 2024-10-04
fix: metadata not showing in results

drew-harris:


drew-harris:
BEFORE: 
image
AFTER:
image

cdxker (approved):
removed a few extra console logs, other than that works great!

PR #2534, opened 2024-10-04
feature: fix search_analytics from returning null when serde serialization fails

cdxker:
We apply a regex pattern to remove chunk_metadata and then add it back in after serialization finishes

### DIFF
--- a/server/src/operators/analytics_operator.rs
+++ b/server/src/operators/analytics_operator.rs
@@ -115,7 +115,7 @@ pub async fn get_search_query(
     search_id: uuid::Uuid,
     clickhouse_client: &clickhouse::Client,
 ) -> Result<SearchQueryEvent, ServiceError> {
-    let clickhouse_query = clickhouse_client
+    let mut clickhouse_query = clickhouse_client
         .query("SELECT ?fields FROM search_queries WHERE id = ? AND dataset_id = ?")
         .bind(search_id)
         .bind(dataset_id)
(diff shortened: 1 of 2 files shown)


server/src/operators/analytics_operator.rs
```diff
@@ -126,7 +126,39 @@ pub async fn get_search_query(
             ServiceError::InternalServerError("Error fetching query".to_string())
         })?;
 
-    let query: SearchQueryEvent = clickhouse_query.into();
+    let re = regex::Regex::new(r#""chunk_html":\s*"(.*?)"\s*,\s*"metadata""#).unwrap();
```
cdxker, line 129: Regex pattern to remove the `chunk_html` body

server/src/operators/analytics_operator.rs
```diff
@@ -126,7 +126,39 @@ pub async fn get_search_query(
...
+    let mut query: SearchQueryEvent = clickhouse_query.into();
+
+    query.results = query
+        .results
+        .iter()
+        .enumerate()
+        .map(|(i, result)| {
+            let mut new_result = result.clone();
+            if !chunk_htmls[i].is_empty() {
+                if let Some(v) = new_result.pointer_mut("/metadata/0/chunk_html") {
```
cdxker, line 155: serde has this cool feature called `pointer_mut` that allows you to edit fine grained without multiple matches.

skeptrunedev (approved):
LGTM

densumesh (approved):

PR #2545, opened 2024-10-08
feature: headless analytics

densumesh:
Still to be done
- [x] make frontends work with headless
- [x] Test all types of headless events
- [x] figure out why query ratings are not serializing/deserializing properlu


cdxker (changes requested):
image

Frontend changes seems to mess up the results view

cdxker (changes requested):
Can we have the user view the custom metadata objects they submit. Use the first 3 fields as the preview and when expanded show the same metadata view we use for the non-custom events

cdxker (changes requested):
This display the correct way for searches but not for `rag_queries`.

cdxker (approved):

PR #2578, opened 2024-10-14
fix: support non-uuid claim subjects

thecaralice:
Previously, the server would expect OpenID Connect subject to be a valid UUID and fail to authenticate the user otherwise. The [OpenID Connect specification][oidc] explicitly allows the subject to be any ASCII string, given that it does not exceed 255 characters.

Was also discussed [on Discord][discord] with @skeptrunedev.

[oidc]: https://openid.net/specs/openid-connect-core-1_0.html#IDToken
[discord]: https://discord.com/channels/1130153053056684123/1293675381370388523


skeptrunedev:
Pinging @densumesh for review on this

densumesh (approved):
Need to test in staging, but local tests seem to work fine

- [x] Login with pre-migration user
- [x] API calls with pre-migration user
- [x] Login with new user created after migration
- [x] API calls with new user created after migration

cdxker (approved):

cdxker:
works great for me, after CI finishes will merge

PR #2609, opened 2024-10-18
feature: add site path to search results

densumesh:


skeptrunedev (changes requested):
Breadcrumbs here should be even less noticeable style wise than the subtext underneath the heading

densumesh:
image

server/src/bin/crawl-worker.rs
```diff
@@ -423,8 +423,12 @@ async fn get_chunks_with_firecrawl(
 
                                     metadata["summary"] = json!(summary.clone());
 
-                                    chunk_html.push_str(format!("\n\n<p>{}</p>", summary).as_str());
+                                    chunk_html.push_str(format!("{}</h2>", summary).as_str());
```
skeptrunedev, line 426: I think we should put both the path and summary in the chunk_html instead of just one or the other

skeptrunedev (dismissed):

cdxker (approved):

PR #2614, opened 2024-10-18
cd/reduce bundle size

cdxker:
- **feature: remove @radix-ui/react-dialog**
- **feature: removed downshift to reduce bundle size**


cdxker:
Closed, merging with #2613

PR #2618, opened 2024-10-18
feature: remove syntax highlighting package to reduce bundle size

cdxker:
Remove syntax highlighting to reduce bundle size


cdxker:
Moving to #2615

PR #2619, opened 2024-10-18
feat: add search component example code

drew-harris:


drew-harris:
image

skeptrunedev (changes requested):
We should include both a react and web component example

drew-harris:
image

skeptrunedev (approved):

PR #2625, opened 2024-10-19
fix(clickhouse): do not force `default` database

thecaralice:
Some code tried to use `default.*` tables even when `CLICKHOUSE_DB` is different.

### DIFF
--- a/server/ch_migrations/chm.toml
+++ b/server/ch_migrations/chm.toml
@@ -1,4 +1,0 @@
-url = "http://localhost:8123"
-user = "clickhouse"
-password = "password"
-database = "default"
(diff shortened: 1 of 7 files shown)


server/ch_migrations/chm.toml
densumesh, line 1: This file is for our `chm` migration tool and needs to be readded

densumesh (changes requested):

server/ch_migrations/chm.toml
thecaralice: It should not be tracked by git as it contains settings specific to the user running `chm`. That's why I added it to `.gitignore`.

densumesh: agreed

densumesh (approved):

Issue #2561, opened 2024-10-11
bugfix: information hovers are not displaying properly


fedhacks:
### Description

These information boxes should be within the bounds of the browser window. They currently are not.

### Target(s)

'dashboard'

<img width="479" alt="Screenshot 2024-10-10 at 8 13 45 PM" src="https://github.com/user-attachments/assets/ccbf40e6-a534-45d8-ab78-865d518c88cf">

### Requirement to close

These information boxes should be within the bounds of the browser window.

### Community channels

Matrix is preferred. Reach out on discord or Matrix for further assistance. 

- [discord](https://discord.gg/CuJVfgZf54)
- [matrix](https://matrix.to/#/#trieve-general:trieve.ai)

drew-harris:
ive always wanted to fix these

PR #2653, opened 2024-10-25
feat: better tooltip component

drew-harris:

Changes tooltip to automatically position itself in the window to not get cut off. 
Fade in animation and optional delay. 
Uses solid portal so its always on top of things


drew-harris:
fixes #2561

PR #2701, opened 2024-11-08
feature: add file chunker microservice

densumesh:


server/src/bin/ingestion-worker.rs
```diff
@@ -416,6 +419,34 @@ pub async fn bulk_upload_chunks(
...
+                .plan
+                .unwrap_or_default()
+                .chunk_count as usize
+        {
+            return Err(ServiceError::BadRequest(
+                "Chunk count exceeds plan limit".to_string(),
+            ));
+        }
+    }
+
```
skeptrunedev, line 449: Why is this getting added in this PR? Were we missing it or was it not being applied ever?

server/src/handlers/file_handler.rs
```diff
@@ -181,6 +182,17 @@ pub async fn upload_file_handler(
...
+
+    create_file_query(
+        file_id,
+        file_size_mb,
+        upload_file_data.clone(),
+        dataset_org_plan_sub.dataset.id,
+        pool.clone(),
+    )
+    .await?;
+
```
skeptrunedev, line 195: Seems odd that this would need to get added? Is there a reason here?

skeptrunedev (changes requested):

skeptrunedev (commented):
- [ ] feature: docker-compose for `file-chunker` with Clickhouse, Minio S3, PDLA, and Redis

server/src/handlers/file_handler.rs
densumesh: Bc before it would've been done in the worker, in this case its being sent to the microservice, so we still need to make sure that the file is created in our PG

server/src/bin/ingestion-worker.rs
densumesh: We were missing it in ingestion worker, this is why we were allowing ppl to have more chubks than they were allowed for site search etc

PR #2717, opened 2024-11-11
enhancement: speed up shopify scraping by pushing chunks into the queue as you go

cdxker:
Send chunks as you go so that the scrape feels snappier when using the Shopify crawling option

### DIFF
--- a/server/src/bin/crawl-worker.rs
+++ b/server/src/bin/crawl-worker.rs
@@ -167,51 +167,6 @@ fn create_chunk_req_payload(
     })
 }
 
-async fn get_chunks_from_shopify(
-    scrape_request: CrawlRequest,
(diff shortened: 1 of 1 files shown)


server/src/bin/crawl-worker.rs
```diff
@@ -599,42 +554,122 @@ async fn crawl(
...
+                }
+            }
+
+            chunk_count += chunks.len();
+            cur_page += 1;
+        }
+
+        (cur_page, chunk_count)
+    } else {
+        // Firecrawl fallback
```
skeptrunedev, line 636:
```suggestion
```

skeptrunedev (changes requested):

skeptrunedev (approved):

PR #2736, opened 2024-11-13
feature: use carousel for product items in chat

vid277:
[Screencast from 11-12-2024 11:18:10 PM.webm](https://github.com/user-attachments/assets/50fe4490-5797-4a08-bb33-eb8b53b52bcc)


clients/search-component/package.json
```diff
@@ -52,6 +52,7 @@
     "@r2wc/react-to-web-component": "^2.0.3",
     "@thumbmarkjs/thumbmarkjs": "^0.14.8",
     "react-markdown": "^9.0.1",
+    "react-snap-carousel": "^0.5.0",
```
skeptrunedev, line 55: How big is this package? Really tempted to say we should build this completely from scratch.

skeptrunedev (changes requested):

clients/search-component/package.json
vid277: This package is ~3.8kb.

skeptrunedev: Sounds good to me then

skeptrunedev (changes requested):
- [ ] carousel should load first while the answer text is still being streamed in 
- [ ] carousel should start open to the first page instead of last

skeptrunedev (approved):
LOOKS GREAT!!

PR #2747, opened 2024-11-15
fix: change shortcut to ctrl-tab

drew-harris:


skeptrunedev (changes requested):
- [ ] release new component on npm
- [ ] update the version in the public page html file

skeptrunedev (approved):

PR #2796, opened 2024-11-19
feature: add route to get message by id

skeptrunedev:
image

### DIFF
--- a/server/src/handlers/message_handler.rs
+++ b/server/src/handlers/message_handler.rs
@@ -13,8 +13,9 @@ use crate::{
         chunk_operator::{get_chunk_metadatas_from_point_ids, get_random_chunk_metadatas_query},
         clickhouse_operator::EventQueue,
         message_operator::{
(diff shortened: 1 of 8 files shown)


server/src/handlers/message_handler.rs
```diff
@@ -132,7 +133,7 @@ pub struct CreateMessageReqPayload {
         ("TR-Dataset" = uuid::Uuid, Header, description = "The dataset id or tracking_id to use for the request. We assume you intend to use an id if the value is a valid uuid."),
     ),
     security(
-        ("ApiKey" = ["readonly"]),
+        ("ApiKey" = ["admin"]),
```
cdxker, line 136: Is this intential? upping the permissions needed to create a message could break UI's

skeptrunedev: It's just for the docs. We have been requiring admin here for a long time, the openapi spec was just off.

cdxker (approved):

PR #2807, opened 2024-11-20
Chunkr integration for pdf2md

akhileshsharma99:
- **added chunkr api as a provider**
- **issues with polling task in clickhouse**
- **chunkr api succesfully implemented**
- **frontend provider choice implemented**
- **chunkr implementation completed**


skeptrunedev (changes requested):
Got the commits cleaned up w/ the merge removed and overall collapsed down to 1 with a `feature: ` prefix. 

One major change off the bat in terms of UI flow:

image

Backend then has an issue when polling for task status: 

image

akhileshsharma99:
Should I just remove the provider logic completely then? Or just change the frontend?

skeptrunedev:
> Should I just remove the provider logic completely then? Or just change the frontend?

Just change the frontend. Also seems like we should re-label the input as `Model/Provider` instead of `Model` once we move `chunkr` there.

pdf2md/server/static/pdf2md.js
```diff
@@ -375,6 +400,8 @@ fileUploadInput.addEventListener("change", (event) => {
           position: { x: "center", y: "top" },
         });
 
+        console.log("data", data);
+
```
skeptrunedev, line 404:
```suggestion
```

skeptrunedev (changes requested):

PR #2817, opened 2024-11-20
bugfix: ecomm group search missing key + query suggestions not always loading on open"

skeptrunedev:
- **v0.2.2**
- **v0.2.3**
- **v0.2.4**
- **bugfix: ecomm group search missing key + query suggestions not always loading on open**


cdxker (changes requested):
Also need to update version here.

https://github.com/devflowinc/trieve/pull/2817/files#diff-22f85ba80bda9fec5935dc305f5e9a327008f7c8a1fc953ada46399d31b79601R67

cdxker (approved):

PR #2822, opened 2024-11-21
enhancement: send an event that scraping has started

cdxker:

### DIFF
--- a/server/src/data/models.rs
+++ b/server/src/data/models.rs
@@ -1781,6 +1781,11 @@ pub enum EventType {
     GroupChunksUpdated { group_id: uuid::Uuid },
     #[display(fmt = "group_chunks_action_failed")]
     GroupChunksActionFailed { group_id: uuid::Uuid, error: String },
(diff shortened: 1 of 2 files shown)


densumesh (approved):

server/src/data/models.rs
```diff
@@ -1782,6 +1782,11 @@ pub enum EventType {
     #[display(fmt = "group_chunks_action_failed")]
     GroupChunksActionFailed { group_id: uuid::Uuid, error: String },
     #[display(fmt = "crawl_completed")]
```
densumesh, line 1784: Needs to be changed

densumesh: Add it to the list on 1805 as well

densumesh (changes requested):

server/src/data/models.rs
cdxker: Fixed and force pushed

densumesh (approved):

PR #2833, opened 2024-11-22
feature: scope api keys to orgs

densumesh:


skeptrunedev:
image
image

Role's aren't getting set correctly when I get the key back after creation

skeptrunedev (changes requested):
- [x] roles aren't getting set correctly on response after API key creation

skeptrunedev (approved):

PR #2834, opened 2024-11-22
bugfix: restore chunk_html to only finding one heading per chunk

skeptrunedev:


server/src/bin/crawl-worker.rs
```diff
@@ -524,7 +524,7 @@ async fn get_chunks_with_firecrawl(
                 metadata["title"] = json!(page_title.clone());
                 metadata["hierarchy"]
                     .as_array_mut()
-                    .unwrap()
+                    .unwrap_or(&mut vec![])
```
skeptrunedev, line 527: WE NEED TO LINT FOR THESE

cdxker (approved):

PR #2837, opened 2024-11-22
feature: flag to disable inserting into postgres on ingest

cdxker:
- **feature: able to insert without postgres, able to get it to search without erroring, need to pull out metadata**
- **feature: extract chunk_html from payload field, need to extract other relevant fileds**


server/src/operators/search_operator.rs
```diff
@@ -1426,15 +1428,34 @@ pub async fn retrieve_chunks_from_point_ids(
...
+            .iter()
+            .map(|search_result| {
+                ChunkMetadataTypes::QdrantChunkMetadata(QdrantChunkMetadata {
+                    // Stopping off right here, we need to convert these into the correct type
+                    qdrant_point_id: search_result.point_id,
+                    chunk_html: search_result.payload.get("content"),
+                    tag_set: search_result.payload.get("tag_set"),
+                    images_urls: search_result.payload.get("images_urls"),
+                    link: search_result.payload.get("link"),
+                    metadata: search_result.payload.get("metadata"),
```
cdxker, line 1447: Currently here right now, shouldn't be too much more work

server/src/operators/search_operator.rs
```diff
@@ -1426,15 +1430,41 @@ pub async fn retrieve_chunks_from_point_ids(
...
+                ChunkMetadataTypes::QdrantChunkMetadata(QdrantChunkMetadata {
+                    qdrant_point_id: search_result.point_id,
+                    chunk_html,
+                    tag_set: None,
+                    images_urls: None,
+                    link: None,
+                    metadata: None,
+                })
+            })
+            .collect()
```
cdxker, line 1459: Stopping off here, but it should be pretty easy to extract the rest of the payload values needed.

densumesh (approved):

PR #2858, opened 2024-11-28
feature: add more dataset usage metrics

densumesh:
image


skeptrunedev (changes requested):
In hindsight, I actually think this is a bad idea. People should call the analytics routes to get this information. Those are more powerful since they already support date ranges which can be used for more advanced billing logic. However, it does seem like we need to add support for getting total number of recommendations performed.

PR #2915, opened 2024-12-08
feature: add hallucination detection to server

densumesh:

### DIFF
--- a/hallucination-detection/examples/rag_truth_test.rs
+++ b/hallucination-detection/examples/rag_truth_test.rs
@@ -131,7 +131,7 @@ async fn run_hallucination_test() -> Result<(), Box<dyn Error>> {
         let start = std::time::Instant::now();
         let hallucination_score = detector
             .detect_hallucinations(&record.response, &[source_info.clone()])
-            .await;
+            .await.unwrap();
         let elapsed = start.elapsed();
         println!("Hallucination detection took: {:?}", elapsed);
 
--- a/hallucination-detection/examples/vectara_test.rs
+++ b/hallucination-detection/examples/vectara_test.rs
@@ -88,7 +88,8 @@ async fn run_hallucination_test() -> Result<(), Box<dyn Error>> {
         let start = std::time::Instant::now();
         let hallucination_score = detector
             .detect_hallucinations(&record.og_sum, &references)
-            .await;
+            .await
+            .unwrap();
         let elapsed = start.elapsed();
         println!("Hallucination detection took: {:?}", elapsed);
 
--- a/hallucination-detection/src/lib.rs
+++ b/hallucination-detection/src/lib.rs
@@ -7,19 +7,30 @@ use std::{
 };
 use tokio::sync::OnceCell;
 
+#[cfg(not(feature = "ner"))]
+#[cfg(feature = "onnx")]
+compile_error!("NER feature must be enabled to use ONNX model");
+
 #[cfg(feature = "ner")]
 use {
     rust_bert::{
-        pipelines::{
-            common::{ModelResource, ModelType, ONNXModelResources},
-            ner::{Entity, NERModel},
-            token_classification::{LabelAggregationOption, TokenClassificationConfig},
-        },
-        resources::RemoteResource,
+        pipelines::ner::{Entity, NERModel},
+        pipelines::token_classification::TokenClassificationConfig,
         RustBertError,
     },
+    std::error::Error,
     std::sync::mpsc,
     tokio::{sync::oneshot, task::JoinHandle},
+};
+
+#[cfg(feature = "onnx")]
+use rust_bert::{
+    pipelines::{
+        common::ONNXModelResources,
+        common::{ModelResource, ModelType},
+        token_classification::LabelAggregationOption,
+    },
+    resources::RemoteResource,
 };
 
 const WORDS_URL: &str =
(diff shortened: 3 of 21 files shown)


hallucination-detection/examples/rag_truth_test.rs
```diff
@@ -131,7 +131,7 @@ async fn run_hallucination_test() -> Result<(), Box<dyn Error>> {
         let start = std::time::Instant::now();
         let hallucination_score = detector
             .detect_hallucinations(&record.response, &[source_info.clone()])
-            .await;
+            .await.unwrap();
```
skeptrunedev, line 134: remove unwrap

hallucination-detection/examples/vectara_test.rs
```diff
@@ -88,7 +88,8 @@ async fn run_hallucination_test() -> Result<(), Box<dyn Error>> {
         let start = std::time::Instant::now();
         let hallucination_score = detector
             .detect_hallucinations(&record.og_sum, &references)
-            .await;
+            .await
+            .unwrap();
```
skeptrunedev, line 92: remove unwrap

hallucination-detection/src/lib.rs
```diff
@@ -381,7 +416,8 @@ mod tests {
 
         let score = detector
             .detect_hallucinations(&llm_output, &references)
-            .await;
+            .await
+            .unwrap();
```
skeptrunedev, line 420: remove unwrap

hallucination-detection/src/lib.rs
```diff
@@ -402,7 +438,8 @@ mod tests {
...
```
skeptrunedev, line 442: remove unwrap

hallucination-detection/src/lib.rs
```diff
@@ -415,13 +452,15 @@ mod tests {
         // Empty input
         let score_empty = detector
             .detect_hallucinations(&String::from(""), &[String::from("")])
-            .await;
+            .await
+            .unwrap();
```
skeptrunedev, line 456: remove unwrap

hallucination-detection/src/lib.rs
```diff
@@ -415,13 +452,15 @@ mod tests {
...
         assert_eq!(score_empty.total_score, 0.0);
 
         // Only numbers
         let score_numbers = detector
             .detect_hallucinations(&String::from("123 456.789"), &[String::from("123 456.789")])
-            .await;
+            .await
+            .unwrap();
```
skeptrunedev, line 463: remove unwrap

hallucination-detection/src/lib.rs
```diff
@@ -430,7 +469,8 @@ mod tests {
                 &String::from("John Smith"),
                 &[String::from("Different Person")],
             )
-            .await;
+            .await
+            .unwrap();
```
skeptrunedev, line 473: remove unwrap

hallucination-detection/src/lib.rs
```diff
@@ -508,7 +548,8 @@ mod tests {
                 &String::from(llm_output),
                 &references.into_iter().map(String::from).collect::<Vec<_>>(),
             )
-            .await;
+            .await
+            .unwrap();
```
skeptrunedev, line 552: remove unwrap

skeptrunedev (changes requested):

server/Dockerfile.server
```diff
@@ -29,11 +29,19 @@ RUN apt-get update -y; \
     ca-certificates \
     curl \
     redis-tools \
+    unzip \
     ;
 
 RUN curl -fsSLO https://github.com/subtrace/subtrace/releases/download/b143/subtrace-linux-amd64 \
     && chmod +x ./subtrace-linux-amd64
 
+RUN curl -L -o libtorch-cxx11-abi-shared-with-deps-2.4.0+cpu.zip https://download.pytorch.org/libtorch/cpu/libtorch-cxx11-abi-shared-with-deps-2.4.0%2Bcpu.zip \
```
cdxker, line 38: This might also need to be located when compiling depenencies.

https://github.com/devflowinc/trieve/pull/2915/files#diff-0092f086be905d516d3457f02246497a74ff3594a5716321fcd13b17b31f9763R15

hallucination-detection/src/lib.rs
densumesh: in test

hallucination-detection/src/lib.rs
densumesh: ``

hallucination-detection/src/lib.rs
densumesh: ``

hallucination-detection/src/lib.rs
densumesh: ``

hallucination-detection/src/lib.rs
densumesh: ``

server/Dockerfile.server
densumesh: The dependency is located using the envs below

cdxker (changes requested):
image

Dockerfile doesn't seem to work, I think we need to include `ort` into the cargo.toml or download libonnxruntime.so some other way

skeptrunedev (changes requested):
- [ ] fix image build
- [ ] show null instead of 0 where hallucination detection has not run 

image

skeptrunedev (changes requested):
This ran fine on the big office computer, but was really slow on my laptop. I'm somewhat concerned about putting this into prod because I think it will necessitate putting the `server` on a GPU node. 

I feel relatively strongly that we need to run the models separately outside of the server.

skeptrunedev:
image

I think the change which got made to old messages w/out hallucination detection didn't appear on the table is also messing up actual hallucinations. Either that or the NER change zeroed out everything? 

Are we just not displaying 0's or do we have a smarter system for not displaying nulls?

skeptrunedev (changes requested):
display detected hallucinations as non-zero on the table

skeptrunedev (changes requested):
image

I also wonder if we should add something like "low confidence queries" and "likely hallucinations" in this PR. 

Feels necessary to be able to sort in descending order by both `top score` and `hallucination score`.

skeptrunedev (approved):
Should be high prio on the ordered queries, but this is good!

PR #2932, opened 2024-12-11
feature: incrementally add pages

densumesh:


skeptrunedev (changes requested):
- [ ] feature: make the upload for a file be a signed PUT url instead of base64
- [ ] feature: allow user to customize prompt on handler (also put in dashboard form)
- [ ] feature: allow user to specify heading_based chunking as an option instead of naive pages (make a group for each page)
- [ ] feature: page and link for each chunk

cdxker (approved):

PR #2935, opened 2024-12-11
fix: trieve-fern-adapter workflow

zobeirhamid:
I notice the `update-trieve-fern-adapter` workflow is not working as the `main` branch is branch protected. Instead of committing to `main`, this will create a new PR with the version bump.


skeptrunedev (approved):
Fix is appreciated!

PR #2949, opened 2024-12-16
feature: add floating action button to docsearch

vid277:
image


skeptrunedev (changes requested):
- [x] position needs to be a prop
- [ ] on/off flag for the floating button needs to be a prop
- [x] readme needs to get new documentation for props
- [ ] example should get new env for testing prop on/off + position

skeptrunedev (approved):

PR #3036, opened 2025-01-02
bugfix: fix suggested queries, and fix window height bug

cdxker:
- **bugfix: fix suggested queries**
- **bugfix: fix window height bug**


skeptrunedev (approved):

fedhacks (approved):
lgtm

PR #3043, opened 2025-01-03
feat: add dataset ID support

gerred:
# Dataset ID Support for MCP Server

Fundamentally, the issue came down to I generated out a non-admin key for just one dataset, and didn't see any datasets, so this fixes that and handles the hierarchy. I imagine there's other rules and should get vitest going for this as well, but then someone additionally will need to make the MCP server testable (i.e. mock out the client and methods).

## Changes
- Added support for specifying a single dataset ID via CLI args (`--dataset-id`) or environment variables (`TRIEVE_DATASET_ID`)
- When dataset ID is provided, the MCP server will only expose and operate on that specific dataset
- Updated documentation to reflect new dataset ID options
- Bumped version to 0.1.3

## Testing
The changes have been tested with both CLI args and environment variables to ensure:
- Server properly initializes with dataset ID
- Search operations are scoped to the specified dataset
- Backwards compatibility is maintained when no dataset ID is provided

## Documentation
README.md has been updated with the new dataset ID configuration options.

### DIFF
--- a/clients/mcp-server/README.md
+++ b/clients/mcp-server/README.md
@@ -27,9 +27,12 @@ Required environment variables:
 - `TRIEVE_API_KEY`: Your Trieve API key from dashboard.trieve.ai
 - `TRIEVE_ORGANIZATION_ID`: Your Trieve organization ID from dashboard.trieve.ai
 
-Alternatively, you can provide these credentials via command-line arguments:
+Optional environment variables:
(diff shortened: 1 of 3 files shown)


skeptrunedev:
@cdxker pinging for review

cdxker (approved):

cdxker:
This looks great, merged!

PR #3055, opened 2025-01-04
bugfix: invitation email redirects to correct organization on dashboard

vid277:
- Also updated the dashboard user context to set the organization based on url query parameters.

image


cdxker:
image

image

Pushed a commit to clean up the html response template. (jinja is so nice)

cdxker (approved):

cdxker (approved):
Works with both existing and new user invites

PR #3067, opened 2025-01-08
feat: autofill favicon public page icon from scrape settings

drew-harris:


drew-harris:
I have a version that runs a fetch() to make sure there actually is a favicon if you want me to add that too.

skeptrunedev (approved):
I was thinking of

PR #3076, opened 2025-01-08
styling: use jsx tailwind for search component styling

drew-harris:


cdxker:
Going to close this and just make a more minified version of tailwind styling. This is too difficult to test.

PR #3121, opened 2025-01-11
bugfix: fix search type typo

cdxker:
- **feature: add tv- prefix to tailwind styles**
- **feature: add inline to `search-component/example`, fix chat with group spacings**
- **bugfix: fix search type typo**


fedhacks (approved):
lgtm

PR #3130, opened 2025-01-13
bugfix: search-component for shopify themes by changing from rem to px with a 16 base

skeptrunedev:


cdxker (approved):
This is great! Thank you so much.

PR #3133, opened 2025-01-13
feature: make query selectors more specific.

cdxker:
- **v0.3.20**
- **feature: make query selectors more specific in case 2 components exist on the same page.**


skeptrunedev (approved):

skeptrunedev (approved):
Amended a small change to fix the lint error for the unused const.

PR #3201, opened 2025-01-18
feature: hideOpenButton param

cdxker:


skeptrunedev (changes requested):
need to add to README

I realize it's already out of date, but let's at least try

skeptrunedev (approved):

PR #3211, opened 2025-01-20
feat: upgrade urllib dep in trieve-py-sdk

skeptrunedev:


drew-harris (approved):
assuming all codegend lgtm

PR #3237, opened 2025-01-24
cleanup: only return header for audio searches

densumesh:


clients/search-component/example/src/routes/index.lazy.tsx
```diff
@@ -154,100 +154,6 @@ export default function Home() {
...
-            use_autocomplete: false,
-            search_type: "fulltext",
-            score_threshold: 0.1,
-          }}
-          floatingButtonPosition={floatingButtonPosition}
-          floatingSearchIconPosition={floatingSearchIconPosition}
-          showFloatingButton={showFloatingButton}
-          showFloatingSearchIcon={showFloatingSearchIcon}
-          showFloatingInput={showFloatingInput}
-          inline={true}
```
skeptrunedev, line 249: Why delete this?

server/src/handlers/chunk_handler.rs
```diff
@@ -2722,17 +2737,22 @@ pub async fn generate_off_chunks(
...
-                last_prev_message
-                    .content
-                    .clone()
-                    .replace("\n", "[NEWLINE]")
-                    .replace("\r", ""),
-            ))
-            .json(completion_content));
+        if data.audio_input.is_some() {
+            return Ok(HttpResponse::Ok()
+                .insert_header(("TR-QueryID", query_id.to_string()))
```
skeptrunedev, line 2742: Should always be returned

clients/search-component/example/src/routes/index.lazy.tsx
densumesh: it messes w/ the example and makes it hard to dev

skeptrunedev (changes requested):

server/src/handlers/chunk_handler.rs
densumesh: its returned in both branches

PR #3251, opened 2025-01-28
feature: add env variable for premium organizations

cdxker:
Added the ability to specify organizations that have much higher priority on ingestion, these will be placed into their separate queue. And ingested by dedicated workers


fedhacks (approved):
Lgtm

PR #3253, opened 2025-01-29
feat: more shopify variables

drew-harris:


drew-harris:
image

cdxker (changes requested):
This looks great, can you remove a few more options, just to simplify it for our users.

- Have suggested queries and follow up queries be on the same setting
- Remove ModalType default it to always be ecommerce
- Can we add in zIndex as a parameter (lots of "fights" happen over z-index)
- Remove the groupSearch option, default it to always groupSearch
- default the search mode to chat
- remove the enable / disable chat parameter

cdxker (approved):

PR #3270, opened 2025-01-31
feat: smooth auth flow and use env variables

drew-harris:


drew-harris:
https://github.com/user-attachments/assets/7b24afa1-cc9f-484d-bd35-dd6f8abdb257

cdxker (approved):

PR #3306, opened 2025-02-04
sk/bugfix shrinking demo page width

skeptrunedev:
- **bugfix: component is shrinking on first message in inline mode**
- **feat: include more options on partner settings for toolbar social icons**

image


fedhacks (approved):
Lgtm

PR #3312, opened 2025-02-04
styling: questions over input

drew-harris:


drew-harris:
made all animations have same quick duration and easing
resorted some of the chat paddings to be more consistent
move the questions after the <copy, thumbsup, thumbsdown> so that when they come in those icons dont fly everywhere

skeptrunedev (approved):

PR #3319, opened 2025-02-05
feature: make crawls a first class feature

densumesh:


server/src/handlers/crawl_handler.rs
```diff
@@ -0,0 +1,184 @@
...
+
+/// Get all crawl requests for a dataset
+///
+/// This endpoint is used to get all crawl requests for a dataset.
+#[utoipa::path(
+    post,
+    path = "/crawl/dataset",
+    context_path = "/api",
+    tag = "Dataset",
+    request_body(content = GetCrawlRequestsReqPayload, description = "JSON request payload to get all crawl requests", content_type = "application/json"),
```
skeptrunedev, line 158: We should not have request bodies for get requests. These should be URL query params.

skeptrunedev (changes requested):

server/src/handlers/crawl_handler.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    Ok(HttpResponse::Ok().json(crawl))
+}
+
+/// Delete a crawl request
+///
+/// This endpoint is used to delete an existing crawl request for a dataset. The request payload should contain the crawl id to delete.
+#[utoipa::path(
+    delete,
+    path = "/crawl/:crawl_id",
```
skeptrunedev, line 117:
```suggestion
    path = "/crawl/{crawl_id}",
```

skeptrunedev: Needs to be in curly braces to fix redoc error

skeptrunedev (changes requested):

skeptrunedev (changes requested):
I think that there's a migration which needs to be `git add`'ed here

frontends/dashboard/src/components/CrawlingSettings.tsx
```diff
@@ -197,20 +184,29 @@ export const CrawlingSettings = () => {
...
+    <>
+      <div
+        class="relative mb-2 flex items-center rounded border border-blue-500 bg-blue-100 px-4 py-3 text-blue-700"
+        role="alert"
+      >
+        <AiOutlineInfoCircle class="mr-1" />
+        <span class="block sm:inline">
+          You can run crawls for multiple sites by submitting multiple crawl
+          requests.
+        </span>
```
skeptrunedev, line 200:
```suggestion
        <span class="block sm:inline">
          You can run crawls for multiple sites by submitting multiple crawl
          requests. See your active crawls in the{" "}
          <a class="underline" href={`/dataset/${datasetId()}/crawl/history`}>
            Crawl History
          </a>{" "}
          tab above.
        </span>
```

skeptrunedev: I think it's good to add a link here so people understand where the history is

skeptrunedev (changes requested):

frontends/dashboard/src/components/CrawlingSettings.tsx
densumesh: its in the tab right above it
image

skeptrunedev: I know, but the link in the info is still worth imo

densumesh: fair

skeptrunedev (approved):

PR #3337, opened 2025-02-06
feat: chat size never shrinks unless reset

drew-harris:
includes:
fix: waits for animation on reset


drew-harris:
https://github.com/user-attachments/assets/bbfde773-2cd5-4e59-81a0-ffc3e34d57a2

cdxker:
I do want to add to this, the chat window *does* shrink. It just takes a very long time for it to hit its max height.

skeptrunedev (approved):

PR #3343, opened 2025-02-08
styling: fix chat scroll popdown

drew-harris:

### DIFF
--- a/clients/search-component/package.json
+++ b/clients/search-component/package.json
@@ -70,6 +70,7 @@
     "@r2wc/react-to-web-component": "^2.0.3",
     "@thumbmarkjs/thumbmarkjs": "^0.16.1",
     "clsx": "^2.1.1",
+    "cva": "npm:class-variance-authority",
     "jotai": "^2.11.0",
     "motion": "^11.17.0",
     "react-intersection-observer-hook": "^3.0.1",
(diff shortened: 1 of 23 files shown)


skeptrunedev (changes requested):
image

Search on mobile does not appear to resize to fullscreen appropriately in ecomm mode

skeptrunedev:
Docs mode desktop search is not expanding large enough

image

skeptrunedev:
Fwiw this is not yet fixing the popdown effect for me on the inline PDP chat, but I guess that's a future issue post-refactor here.

skeptrunedev (changes requested):
image

This should only be fullscreen'd for mobile. We need to preserve the existing width limitations we have had for desktop

skeptrunedev (changes requested):
image

Component still has no space from the top and doesn't expand to full screen

skeptrunedev:
image

Ideal mobile view

skeptrunedev (changes requested):
- [ ] Maximum width in desktop view of the popover should be `1440px` including `px-[16px]`. 
- [ ] In desktop mode, we should space off the top to the same distance as we are currently spacing off the sides

image

skeptrunedev (changes requested):
image

Max height for both `search` and `chat` should be the same

drew-harris:
https://github.com/user-attachments/assets/c095c777-02a4-4053-b7a5-b6943196d549

skeptrunedev (approved):

PR #3367, opened 2025-02-13
bugfix: use title instead of chunkHtml when estimating title

cdxker:


server/src/bin/crawl-worker.rs
```diff
@@ -83,22 +83,6 @@ fn create_shopify_chunk_req_payload(
...
-            product_title = product_title.replace(remove_string, "");
-            variant_title = variant_title.replace(remove_string, "");
-        });
-    }
-    if let Some(body_remove_strings) = &scrape_request.crawl_options.body_remove_strings {
-        body_remove_strings.iter().for_each(|remove_string| {
-            product_body_html = product_body_html.replace(remove_string, "");
-        });
-    }
-
```
skeptrunedev, line 101: I'm confused why these would run given `heading_remove_strings` or `body_remove_strings` would have to be specified for the inside of the if to trigger.

skeptrunedev (changes requested):

server/src/bin/crawl-worker.rs
cdxker: This is just a no-op.  removed it because its just confuisng

skeptrunedev (approved):

PR #3381, opened 2025-02-15
feature: add group search with fuzzy matching to groups tab

vid277:
image
image

### DIFF
--- a/frontends/search/src/components/BookmarkPopover.tsx
+++ b/frontends/search/src/components/BookmarkPopover.tsx
@@ -6,6 +6,7 @@ import {
   createEffect,
   createSignal,
   useContext,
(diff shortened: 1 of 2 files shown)


densumesh (changes requested):
I think we should do this with just a substring match rather than fuzzy matching. less dependencies is always better

skeptrunedev:
> I think we should do this with just a substring match rather than fuzzy matching. less dependencies is always better

I think it's fine. Reasonable lib to add

skeptrunedev (approved):

skeptrunedev (approved):
LGTM, but I wish we kept fuzzy and I added name + tracking id + description to the search and table

PR #3387, opened 2025-02-18
feature(shopify): default to only supporting one dataset.

cdxker:
image

image

This setup only works well if you choose an org with 1 dataset.
I have a work in progress for the ability to change your dataset, got blocked by the way remix does post requests.


clients/trieve-shopify-extension/Dockerfile
```diff
@@ -18,4 +18,8 @@ COPY . .
 
 RUN npm run build
 
+# RUN curl -fsSLO https://github.com/subtrace/subtrace/releases/download/b257/subtrace-linux-amd64 \
+#     && chmod +x ./subtrace-linux-amd64
+
+# CMD ["./subtrace-linux-amd64", "run", "-tls=true", "--", "npm", "run", "docker-start"]
```
densumesh, line 24: did we remove subtrace for a reason?

densumesh (approved):

PR #3401, opened 2025-02-20
fix: background scroll

drew-harris:


drew-harris:
https://github.com/user-attachments/assets/936fe3d4-6b6c-4dfb-a545-411e8ab1ef19

skeptrunedev (changes requested):
This does not fix the problem for `inline` mode. However, it does for the overall modal.

PR #3417, opened 2025-02-22
feature: listen to shopify product webhooks

densumesh:
### testing
- first migrate db `npx prisma migrate dev`
- run shopify app `yarn dev`
- deploy app for webhooks `npx shopify app deploy`
- Go to shopify app page and it will auto create a new dataset 
- run a crawl to save the settings
- Go to product and create/update a product and you should see the webhooks being called and that reflected in the search index


cdxker (changes requested):
Can we also save the state metafields used here.

skeptrunedev (dismissed):

skeptrunedev (approved):

PR #3421, opened 2025-02-23
feature: batch upvotes and downvotes

densumesh:

### DIFF
--- a/frontends/shared/types.ts
+++ b/frontends/shared/types.ts
@@ -461,7 +461,7 @@ export interface AnalyticsFilter {
   search_type?:
     | "search"
     | "autocomplete"
-    | "rag"
+    | "rag_chunks"
     | "search_over_groups"
     | "search_within_groups";
 }
--- a/server/ch_migrations/1740267077_migrate_rag_queries_to_replacing_merge_tree/up.sql
+++ b/server/ch_migrations/1740267077_migrate_rag_queries_to_replacing_merge_tree/up.sql
@@ -1,0 +1,31 @@
+CREATE TABLE rag_queries_new
+(
+    `id` UUID,
+    `rag_type` String,
+    `user_message` String,
+    `search_id` UUID,
+    `results` Array(String),
+    `llm_response` String,
+    `dataset_id` UUID,
+    `created_at` DateTime,
+    `user_id` String,
+    `query_rating` String DEFAULT 0,
+    `json_results` Array(String),
+    `hallucination_score` Float64 DEFAULT 0.,
+    `detected_hallucinations` Array(String) DEFAULT [],
+    `top_score` Int32 DEFAULT 0
+)
+ENGINE = ReplacingMergeTree()
+PARTITION BY (toYYYYMM(created_at), dataset_id)
+ORDER BY (id, created_at)
+SETTINGS index_granularity = 8192;
+
+
+INSERT INTO rag_queries_new
+SELECT * FROM rag_queries;
+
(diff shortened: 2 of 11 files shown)


server/src/operators/clickhouse_operator.rs
```diff
@@ -112,27 +98,115 @@ pub async fn send_to_clickhouse(
...
+                            _ => None,
+                        })
+                        .find(|e| e.id == event.query_id);
+                }
+
+                let mut rag_event = if let Some(event) = rag_event {
+                    event
+                } else {
+                    continue;
+                };
```
densumesh, line 143: I know this looks a little sus, but it is probably the best way to do it bc it ensures that CH never will face heavy load bc of updates.

frontends/shared/types.ts
```diff
@@ -461,7 +461,7 @@ export interface AnalyticsFilter {
   search_type?:
     | "search"
     | "autocomplete"
-    | "rag"
+    | "rag_chunks"
```
skeptrunedev, line 464: Is this going to cause users currently using the API to have to change something?

skeptrunedev (changes requested):

frontends/shared/types.ts
densumesh: No it's not I just set it wrong when I first made it

skeptrunedev (approved):

server/ch_migrations/1740267077_migrate_rag_queries_to_replacing_merge_tree/up.sql
```diff
@@ -0,0 +1,31 @@
...
+
+
+INSERT INTO rag_queries_new
+SELECT * FROM rag_queries;
+
+RENAME TABLE 
+    rag_queries TO rag_queries_old,
+    rag_queries_new TO rag_queries;
+
+DROP TABLE rag_queries_old;
```
skeptrunedev, line 31: This seems like a bad idea? Doesn't it mean we would lose all previous data?

skeptrunedev (changes requested):

server/ch_migrations/1740267077_migrate_rag_queries_to_replacing_merge_tree/up.sql
cdxker: It won't lose all previous data, it makes a copy of the database and drops the old. iirc he mentioned the table was 180kb. 

I would want to double check the size of this on other on-prem clusters though.

densumesh: I think the best way to go about this is the port forward the database and run the migration locally to prevent it from blocking server from starting

server/ch_migrations/1740267077_migrate_rag_queries_to_replacing_merge_tree/up.sql
```diff
@@ -0,0 +1,31 @@
...
+    `top_score` Int32 DEFAULT 0
+)
+ENGINE = ReplacingMergeTree()
+PARTITION BY (toYYYYMM(created_at), dataset_id)
+ORDER BY (id, created_at)
+SETTINGS index_granularity = 8192;
+
+
+INSERT INTO rag_queries_new
+SELECT * FROM rag_queries;
```
skeptrunedev, line 25: Nice! Surprised I missed this

skeptrunedev (approved):

PR #3432, opened 2025-02-25
feature: track component clicks and chats by component name

cdxker:
Copy of #3429

Requesting changes, this seems to cause a react load error on initial mount

image


skeptrunedev (changes requested):
image

image

Analytics event when I open the component seems to error

skeptrunedev (approved):
image

Will not block merge, but it would be nice if I could see the details of the event here for which variant of the component got opened.

PR #3434, opened 2025-02-26
feature: disable group search and specify a filter

densumesh:


skeptrunedev (approved):
Tested and doesn't seem to cause any harm to existing component

PR #3440, opened 2025-02-26
feature(shopify): split variant name by spaces

cdxker:
Split `variantTitle` by " / " in order to extract all the variant titles. Should help for more accurate filtering.


fedhacks (approved):
lgtm

PR #3443, opened 2025-02-26
feature: add query rating filters

densumesh:

https://github.com/user-attachments/assets/c41dae5f-d188-4e0c-b7ae-1f00fa74a081


skeptrunedev (changes requested):
I know that our component is `thumbs up/down`, however given that our actual API is any rating, I think we should make this filter a `> | < ` situation instead of combobox for up/down.

densumesh:
image

skeptrunedev (approved):

PR #3476, opened 2025-03-05
feat: shopify analytics charts

drew-harris:


drew-harris:
image

densumesh:
Can we move the date range to the right side and make the table pagination buttons bg the same color as the card + move them to the right side of the card as well

densumesh (changes requested):
Can we move the date range to the right side and make the table pagination buttons bg the same color as the card + move them to the right side of the card as well. Also is it possible to get rid of the dots on the button and just keep the tooltips, feel like itd look cleaneer

densumesh (approved):

PR #3488, opened 2025-03-06
fix: chat height

drew-harris:


drew-harris:
https://github.com/user-attachments/assets/26bf46ba-22ad-49cd-9943-554732dede92

skeptrunedev (approved):

PR #3502, opened 2025-03-09
feature: add fingerprint, url, and component name in each event

densumesh:


skeptrunedev (changes requested):
I think we should add a JSON field called componentProps` and just track everything. It can't hurt to have more data in the long term.

clients/search-component/src/TrieveModal/index.tsx
```diff
@@ -107,6 +108,36 @@ const Modal = () => {
...
+        event_type: "view",
+        items: [],
+        metadata: {
+          component_name: props.componentName ?? "trieve-modal-search",
+          page_url: window.location.href,
+          component_props: props,
+          fingerprint
+        },
+      });
+    }, 1);
```
skeptrunedev, line 133: Why is this in a setTimeout for `1ms`?

skeptrunedev (changes requested):

skeptrunedev (changes requested):
All places were `sendCtrData` is called should also have gotten these changes

skeptrunedev (approved):

PR #3506, opened 2025-03-11
feat: scroll modal down on follow up questions

drew-harris:


clients/search-component/src/utils/hooks/useSizing.ts
```diff
@@ -27,7 +27,7 @@ const popupClass = cva(
       },
       mode: {
         search: [],
-        chat: [],
+        chat: ["md:!tv-max-h-[60vh]"],
```
drew-harris, line 30: definitely down to personal preference but I think with the auto scroll, it feels better when its not as tall.

drew-harris:
https://github.com/user-attachments/assets/8647e203-83b5-449f-bd6e-1c7fb6a39be7

skeptrunedev (changes requested):
image

Was this tested via pointing a demo page at local? 

I tried and both the 60% height limit along with the auto scroll did not seem to be getting applied. I checked to make sure I was both using the latest code on this branch along with local component for the demo page I tested before requesting changes. 

Seems like there is an issue causing the changes to not be applied.

skeptrunedev (approved):

PR #3629, opened 2025-03-27
feature(component): chat history view, feature(shopify): add chat history view to shopify store

cdxker:
This pr adds a parameter to the `search-component` for `previewTopicId`. This puts the component into "preview" mode and renders the previous topic thread.

It also adds a preview component into the shopify store.


fedhacks (approved):
lgtm

PR #3651, opened 2025-03-31
bugfix: make jina default embedding model, feature: auto log in if only 1 org

densumesh:
## Description\n\nbugfix: make jina default embedding model, feature: auto log in if only 1 org\n\n## Changes\n\n- Added features as described above

### DIFF
--- a/server/src/data/models.rs
+++ b/server/src/data/models.rs
@@ -2448,7 +2448,7 @@ pub struct DatasetEventCount {
     "server_configuration": {
         "LLM_BASE_URL": "https://api.openai.com/v1",
         "EMBEDDING_BASE_URL": "https://api.openai.com/v1",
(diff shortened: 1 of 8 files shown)


server/src/data/models.rs
```diff
@@ -2448,7 +2448,7 @@ pub struct DatasetEventCount {
     "server_configuration": {
         "LLM_BASE_URL": "https://api.openai.com/v1",
         "EMBEDDING_BASE_URL": "https://api.openai.com/v1",
```
skeptrunedev, line 2450: I think you missed `EMBEDDING_BASE_URL` in a few places and also didn't change the `DIMENSION` whatever that is called

skeptrunedev (changes requested):

skeptrunedev (approved):
LGTM

PR #3661, opened 2025-04-01
feature(shopify): add followup query analytics

vid277:
image


clients/search-component/src/TrieveModal/Chat/SuggestedQuestions.tsx
```diff
@@ -16,13 +16,45 @@ export const SuggestedQuestions = ({
...
+    const lastMessage = JSON.parse(
+      window.localStorage.getItem("lastMessage") ?? "{}",
+    );
+
+    const requestId =
+      Object.keys(lastMessage)[0] || "00000000-0000-0000-0000-000000000000";
+
+    await trieveSDK.sendAnalyticsEvent({
+      event_name: `site-followup_query`,
+      event_type: "followup_query",
```
densumesh, line 38: Don't create a new event_type just use the click event with that name

clients/search-component/src/TrieveModal/Chat/SuggestedQuestions.tsx
```diff
@@ -16,13 +16,45 @@ export const SuggestedQuestions = ({
...
+
+    const lastMessage = JSON.parse(
+      window.localStorage.getItem("lastMessage") ?? "{}",
+    );
+
+    const requestId =
+      Object.keys(lastMessage)[0] || "00000000-0000-0000-0000-000000000000";
+
+    await trieveSDK.sendAnalyticsEvent({
+      event_name: `site-followup_query`,
```
skeptrunedev, line 37: I believe that either this one or the one before is not a followup but rather an initial suggestion?

skeptrunedev (changes requested):

densumesh (changes requested):

clients/trieve-shopify-extension/app/components/analytics/chat/PopularSuggestedQueries.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const { data } = useQuery(
+    popularSuggestedQueriesQuery(trieve, filters, page),
+  );
+
+  const client = useQueryClient();
+  useEffect(() => {
+    client.prefetchQuery(
+      popularSuggestedQueriesQuery(trieve, filters, page + 1),
+    );
+  }, [page]);
```
drew-harris, line 24: filters should be in the dependency array

clients/search-component/src/TrieveModal/Chat/SuggestedQuestions.tsx
```diff
@@ -23,6 +23,14 @@ export const SuggestedQuestions = ({
     return null;
   }
 
+  const handleSuggestedQuestion = async (q: string) => {
```
skeptrunedev, line 26: This needs to be a `useCallback` so it doesn't trigger re-renders inappropriately

skeptrunedev (changes requested):

clients/search-component/src/TrieveModal/Chat/SuggestedQuestions.tsx
densumesh: I think we should track these too

densumesh (approved):

PR #3662, opened 2025-04-01
feat: redesign shopify onboarding and add more control to component install

drew-harris:
https://github.com/user-attachments/assets/41ebdb1c-20d9-4d1b-8025-6e7034420db1


skeptrunedev (changes requested):
image

skeptrunedev (changes requested):
image

This is spinning every time I refresh even when the step has been completed before

skeptrunedev:
> image
> 
> This is spinning every time I refresh even when the step has been completed before

Merging w/out fix because it's incoming already

skeptrunedev (approved):

PR #3705, opened 2025-04-07
feat(trieve-shopify-extension): add default search mode and default search filters for global and inline component

skeptrunedev:


cdxker (changes requested):
Need to also add it for chatFilters

skeptrunedev:
> Need to also add it for chatFilters

Fixed. Good catch, I had no idea that was even a prop

cdxker (approved):

PR #3707, opened 2025-04-07
feature: add dittofeed events to users that are on shopify

cdxker:
Added a route on the server `/api/shopify/user_event` for the remix app to send generic dittofeed tracking events to the server. As well as updated the shopify app to use the new route.

Events Being sent:

```
shopify_linked { (only one with extra metadata)
    store_name: string; (organization_name is the likely just store_name)
    organization_id: string;
}
```

- `catalogue_indexed`
- `fist_chat_completed`
- `global_component_added`
- `pdp_component_added`


clients/trieve-shopify-extension/app/routes/app.setup.tsx
```diff
@@ -225,15 +193,43 @@ export const loader = async (args: LoaderFunctionArgs) => {
...
     key,
     fetcher,
-  );
+    session,
+    crawlOptions,
+  ).catch(console.error);
 
   trieve.datasetId = datasetId;
-
+  console.log("redirecting to app!");
```
skeptrunedev, line 233:
```suggestion
```

server/src/data/models.rs
```diff
@@ -3910,7 +3910,7 @@ impl Default for StripePlan {
             chunk_count: 1000,
             file_storage: 1024 * 1024,
             user_count: 5,
-            dataset_count: 10,
+            dataset_count: 300,
```
skeptrunedev, line 3913:
```suggestion
            dataset_count: 10,
```

skeptrunedev (changes requested):

skeptrunedev (approved):

PR #3740, opened 2025-04-12
feature: ability to create api only users if a given secret key is provided

cdxker:
Route to allow for api only users to be created only if a secret key is passed between both the client and server. This also will associate a user by email if they eventually signup through oidc.


clients/trieve-shopify-extension/README.md
```diff
@@ -350,10 +350,6 @@ This template uses [Remix](https://remix.run). The following Shopify tools are a
 
 ![photo-guide](https://cdn.trieve.ai/github/shopify-app-extension-uuid-guide.png)
 
-## Things to know about `REMIX_SERVER_URL`
```
skeptrunedev, line 353: Why remove this?

clients/trieve-shopify-extension/app/auth.ts
```diff
@@ -1,51 +1,73 @@
...
+        }
+      }
+    `);
+
+    const queryResponse = await query?.json();
+
+    const shop_name = queryResponse?.data?.shop?.name ?? "";
+    const shop_email = queryResponse?.data?.shop?.email ?? "";
+
+    const userCredentialsResponse = await fetch(`${getTrieveBaseUrlEnv()}/api/auth/create_api_only_user`, {
```
skeptrunedev, line 35: You could use the SDK, but not a big deal

clients/trieve-shopify-extension/app/components/MustLoginPage.tsx
```diff
@@ -1,42 +0,0 @@
-import { Box, CalloutCard } from "@shopify/polaris";
```
skeptrunedev, line 1: Feels really good to delete this fully

clients/trieve-shopify-extension/prisma/schema.prisma
```diff
@@ -33,14 +33,11 @@ model Session {
 
 model ApiKey {
   id        String   @id @default(uuid())
-  userId    String
   organizationId String @default("")
   currentDatasetId String? @db.Uuid
-  shop      String @default("")
+  shop      String @default("") @unique
```
skeptrunedev, line 38: Is this guaranteed to be unique from Shopify?

server/src/handlers/auth_handler.rs
```diff
@@ -631,3 +672,92 @@ pub async fn login_cli() -> Result<HttpResponse, ServiceError> {
...
+        .to_str()
+        .map_err(|_| {
+            ServiceError::BadRequest("Failed to parse X-Shopify-Authorization header".to_string())
+        })?;
+
+    if access_key != get_env!("SHOPIFY_SECRET_KEY", "SHOPIFY_SECRET_KEY must be set") {
+        return Err(ServiceError::BadRequest(
+            "Invalid session token".to_string(),
+        ));
+    }
```
skeptrunedev, line 722: You could implement this as middleware, but it's fine to wait and do it later the next time we are doing something that needs this shopify header.

clients/trieve-shopify-extension/prisma/schema.prisma
cdxker: shop name is a domain name. By nature of it being a domain name it is unique.

skeptrunedev (approved):

PR #3741, opened 2025-04-12
feature: add settings item for setting the checkout selector

densumesh:
image


skeptrunedev (approved):
Worked perfectly. I'm also generally happy that we have an established pattern for these other appMetafields now.

PR #3763, opened 2025-04-15
feat: highlight add to cart events in search component view

drew-harris:
image


skeptrunedev (changes requested):
Definitely a nit, but I really feel like this should be green instead of red.

cdxker (approved):

PR #3776, opened 2025-04-17
feat: connect shopify app with judge.me

drew-harris:


cdxker (changes requested):
image

Seems that the user is always logged in here.

cdxker (approved):
This works!

PR #3790, opened 2025-04-17
feature: use checkout end events for purchase

densumesh:


cdxker (changes requested):
Can you make sure the shopify frontend doesn't look for `site-checkout` event types and is `site-checkout_end` instead

cdxker (approved):

PR #3794, opened 2025-04-18
feat: "update" component on theme & make checks more self-explanatory

drew-harris:


drew-harris:
https://github.com/user-attachments/assets/1a3c4d22-4aa5-4495-b494-ebd830d99727

cdxker (changes requested):
It works for the pdp chat, but it doesn't work for the global Component. Can you figure out why?

GlobalNotChecking

cdxker (approved):
It works!

PR #3822, opened 2025-04-25
feature: req-reply cache for get tool function params

densumesh:

https://github.com/user-attachments/assets/8014e7fc-afbc-4364-886b-4ae19652f7b3

### DIFF
--- a/server/Cargo.lock
+++ b/server/Cargo.lock
@@ -84,23 +84,23 @@ dependencies = [
 
 [[package]]
 name = "actix-http"
-version = "3.8.0"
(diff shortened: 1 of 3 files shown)


cdxker (changes requested):
The cache doesn't seem to get refreshed if used within the time span. View the screenshot below.

Reproduced by setting ttl to 5 seconds. 

RefreshCache

See the timings from the right.
image

cdxker (approved):
The requested change will be converted to an issue at https://github.com/densumesh/actix-request-reply-cache

PR #3827, opened 2025-04-29
feature: added ability to scope users to api routes

densumesh:


skeptrunedev (changes requested):
https://github.com/user-attachments/assets/eeb9de77-7700-4c60-9eba-368a3e0b0a57

skeptrunedev (approved):

PR #3835, opened 2025-05-02
adds e2e tests + evals

mclenhard:
Adds new e2e test that loads an MCP client, which in turn runs the server and processes the actual tool call. Afterwards, it then grades the response for correctness. 

note: I'm the package author


skeptrunedev (approved):
Hey @mclenhard! Thanks for the contribution. Some notes to flag: 

- `searchEval` is a `const` that was defined twice in the evals.ts file 
- our mcp server was in `clients/mcp-server` and we only have the root package.json to coordinate workspaces. I feel like it's reasonable that you could have looked there and cleaned up the PR before submitting. 

Regardless, I fixed those issues in another commit - https://github.com/devflowinc/trieve/pull/3835/commits/97cf574a85fdfa6e0037f38366ab4a8b64fc5379 - and this is good to merge! Thanks for contributing.

PR #3842, opened 2025-05-03
cleanup: rewrite collapse script in go w/ concurrency

densumesh:


docker/collapse-query-script/collapse_queries.go
```diff
@@ -0,0 +1,465 @@
...
+			}(dup)
+		}
+
+		wg.Wait()
+		close(errCh)
+
+		for err := range errCh {
+			tx.Rollback()
+			return err
+		}
```
skeptrunedev, line 379: I think this is right, but want to make sure you thought about it. Not sure, but seems like the behavior here is that if one of the `batchSize`  exec's fails, you roll all of them back. That **sounds** correct, but verifying.

docker/collapse-query-script/collapse_queries.go
```diff
@@ -0,0 +1,465 @@
...
+	return nil
+}
+
+func processDataset(conn *sql.DB, datasetID uuid.UUID) {
+	log.Printf("Processing dataset %s", datasetID)
+
+	// Get last collapsed timestamp
+	lastCollapsed, exists, err := getDatasetLastCollapsed(conn, datasetID)
+	if err != nil {
+		log.Printf("Error getting last collapsed for dataset %s: %v", datasetID, err)
```
skeptrunedev, line 395: Same comment about bubbling this up to the caller

skeptrunedev (changes requested):

docker/collapse-query-script/collapse_queries.go
densumesh: yeah we should probably roll all of them back as the non duplicate query could be across the batch boundry so this removes any edge cases from that

skeptrunedev (approved):

PR #3889, opened 2025-05-14
bugfix: table scroll issues + feature: allow for setting params of price tool call

skeptrunedev:
- **feature(server+dashboard+component): add component prop for PriceTool parameters + necessary server and dashboard code for allowing it to be set on demo page**
- **bugfix(search-component): enable horizontal scroll on tables within the component**


fedhacks (dismissed):
lgtm

PR #3892, opened 2025-05-14
bugfix: search component pdp fulltext search does not always get results

skeptrunedev:
- **bugfix(search-component): use both search and getChunks for the inline PDP chat so we always get chunks**
- **feat(search-component): v0.4.74**


fedhacks (approved):
lgtm

PR #3900, opened 2025-05-16
feature(n8n-nodes-trieve): n8n trieve create chunk node /api/chunk/

cdxker:


clients/n8n-nodes-trieve/README.md
cdxker, line 1: I also included how setup instructions.

skeptrunedev (approved):

PR #3922, opened 2025-05-20
cleanup(component+dashboard): remove + 1 on length for loading + change all SERP page to just SERP

skeptrunedev:


fedhacks (approved):
lgtm

PR #3939, opened 2025-05-21
feature: try on items + image upload zone

vid277:
image
image


skeptrunedev (changes requested):
image

Once the image is uploaded and sent in a message, it needs to be removed from the input relative pos

skeptrunedev (changes requested):
image

This should default to not displaying if `imageStarterText` is not defined. 

- [ ] feature(dashboard): user needs to be able to specify `imageStarterText` in their advanced options

cdxker (approved):

cdxker (approved):

PR #3964, opened 2025-05-23
feature: allow for image upload with default questions

vid277:


cdxker (approved):

cdxker (dismissed):
Seems like there's a bug with a state race condition

skeptrunedev (approved):

PR #3976, opened 2025-05-24
feature: pdp image upload and generation

vid277:


vid277:
image

cdxker (changes requested):
Should also remove the other package that was being used before. Not sure what this is meant to fix.

cdxker (approved):

PR #3991, opened 2025-05-29
feature: improve default suggested user questions

vid277:


cdxker (approved):

cdxker (changes requested):
This should only be applied for ecommerce only modes. Can you add a flag for that.

PR #4024, opened 2025-09-10
sec: bump permission on creating an `organization_api_key`

cdxker:

The ability to create organization api keys should only be allowed for Admin users
or greater. This is a security feature to prevent generating arbitrary keys from an unscoped
read-only user api key.


densumesh:
You're welcome lmao

